PR #3, opened 2023-04-04
feat: implement draggable thumbnail overlay for minimized active call

SERDUN:


lib/features/call/observers/overlay_navigator_observer.dart
ycherniavskyi, line 1: In general, I am not sure that this is the correct/necessary class. `RouteObserver` already has the functionality to subscribe/unsubscribe and instead of using a bit odd `Function()?` callbacks class that interested in route changes just needs to utilize already existing mixin [`RouteAware`](https://api.flutter.dev/flutter/widgets/RouteAware-class.html).

lib/features/call/widgets/call_overlay.dart
```diff
@@ -0,0 +1,94 @@
...
+  void initState() {
+    super.initState();
+    widget.observer.setPopListener(_showCallOverlay);
+    widget.observer.setPushListener(_overlayController.disposeAll);
+
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      final appBarTheme = AppBarTheme.of(context);
+      final screenInfo = MediaQuery.of(context);
+
+      const callOverlayEntityMargin = Offset(8, 8);
```
ycherniavskyi, line 43: Move this hardcoded variable to widget's default parameter value.

lib/features/call/widgets/call_overlay_dialog.dart
```diff
@@ -0,0 +1,123 @@
...
+    required this.remoteStream,
+    required this.isVideoCall,
+    required this.onTap,
+  });
+
+  final DateTime createdTime;
+  final MediaStream? remoteStream;
+  final String name;
+  final bool isVideoCall;
+  final Function() onTap;
```
ycherniavskyi, line 22: When using a generally adapted parameter name, such as `onTap`, please leave the original type - `GestureTapCallback onTap`.

lib/features/call/widgets/call_overlay_dialog_minimized.dart
```diff
@@ -0,0 +1,127 @@
...
+    required this.remoteStream,
+    required this.isVideoCall,
+    required this.onTap,
+  });
+
+  final DateTime createdTime;
+  final MediaStream? remoteStream;
+  final String name;
+  final bool isVideoCall;
+  final Function() onTap;
```
ycherniavskyi, line 23: The same as in the previous comment - please use `GestureTapCallback`.

packages/webtrit_overlay_floating/lib/src/overlay_widget.dart
```diff
@@ -0,0 +1,89 @@
+import 'package:flutter/widgets.dart';
+
+import 'overlay_entry_data.dart';
+
+class OverlayController {
+  Function(String id, OverlayEntryData overlayEntryData)? overlayAddCallback;
+  Function(String id)? overlayRemoveCallback;
+  Function()? overlayDisposeAllCallback;
```
ycherniavskyi, line 8: I am not sure that using a public parameter is not the antipattern. Is not better to use at last setters and getters?

ycherniavskyi (changes requested):
I propose to rename a bit long package name `webtrit_overlay_floating` to something like `floating_overlay` or `draggable_overlay`.

ycherniavskyi (approved):

PR #4, opened 2023-04-04
feat: add back button to active call screen

SERDUN:


ycherniavskyi:
@SERDUN I propose trying to utilize the standard material `AppBar` with:
* transparent background
* `title` for `call_description_*`
* `bottom` for other call info texts

Another proposal is to leave the current layout as it is, but add a transparent `AppBar` with a back button via `Stack`.

PR #6, opened 2023-05-08
fix: add event to create a connection on web platform

SERDUN:


SERDUN:
For web in mixin WidgetsBindingObserver  method don't call didChangeAppLifecycleState.
And becouse i initialize our connection in init state.

DidChangeAppLifecycleState - not supported on web

ycherniavskyi:
@SERDUN, after a bit of digging, I've found that even on mobile, the signaling connection starts not due to _AppLifecycleStateChanged, but because of _ConnectivityResultChanged. Consequently, it seems more logical to initiate the signaling connection based on the app's connectivity state, rather than its lifecycle state. So, I'm planning on making a few tweaks to your PR. 🙂

PR #7, opened 2023-05-08
feat: add platform check for permissions

SERDUN:


SERDUN:
Plugin which contains class Helper https://pub.dev/packages/flutter_webrtc don't have web support

lib/features/call/bloc/call_bloc.dart
```diff
@@ -1331,8 +1332,9 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
           : false,
     };
     final localStream = await navigator.mediaDevices.getUserMedia(mediaConstraints);
-    await Helper.setSpeakerphoneOn(video);
-
+    if (PlatformInfo().isMobile) {
```
ycherniavskyi, line 1335: It seems that `kIsWeb` could be used here.

PR #8, opened 2023-05-12
fix: add parameter for allowing  use inherited media query

SERDUN:


screenshots/lib/widgets/screenshot_app.dart
```diff
@@ -42,6 +42,7 @@ class ScreenshotApp extends StatelessWidget {
                 darkTheme: themeProvider.dark(),
                 debugShowCheckedModeBanner: false,
                 routerDelegate: ScreenshotRouterDelegate(child),
+                useInheritedMediaQuery: true,
```
ycherniavskyi, line 45: It appears that the [`useInheritedMediaQuery` property](https://api.flutter.dev/flutter/material/MaterialApp/useInheritedMediaQuery.html) has been marked as deprecated. Given our recent Flutter SDK update, do we still need to include it, or should we seek an alternative solution?

([Reference to the actual deprecation](https://github.com/flutter/flutter/blob/f18cb2d064fd60b3d154bf07f55eb2aa0c111278/packages/flutter/lib/src/material/app.dart#L249C7-L249C7) in the source code for historical context, as the documentation will eventually be updated and this property will be removed.)

ycherniavskyi (changes requested):

PR #9, opened 2023-05-12
fix: add limit for keypad size

SERDUN:
Added limit for keypads items, it's fix most not covered screen size

### DIFF
--- a/lib/features/keypad/widgets/keypad.dart
+++ b/lib/features/keypad/widgets/keypad.dart
@@ -2,7 +2,6 @@ import 'dart:math';
 
 import 'package:flutter/material.dart';
 
-import 'package:webtrit_phone/app/constants.dart';
 import 'package:webtrit_phone/models/models.dart';
 import 'package:webtrit_phone/theme/theme.dart';
 import 'package:webtrit_phone/widgets/widgets.dart';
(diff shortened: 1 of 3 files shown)


lib/features/keypad/widgets/keypad.dart
```diff
@@ -1,24 +1,28 @@
+import 'dart:math';
+
 import 'package:flutter/material.dart';
 
 import 'package:webtrit_phone/models/models.dart';
 import 'package:webtrit_phone/theme/theme.dart';
 import 'package:webtrit_phone/widgets/widgets.dart';
 
 class Keypad extends StatelessWidget {
+  final void Function(String) onKeypadPressed;
```
ycherniavskyi, line 10: Could you please clarify the rationale for moving this variable before the constructor?
If you're aiming for consistency with `Actionpad`, then our current style guide actually suggests placing instance variables after the constructor in `Actionpad`. Could you consider making this change for alignment with our coding conventions?

lib/features/keypad/widgets/keypad.dart
```diff
@@ -1,24 +1,28 @@
...
+
+  final double _maxSizeLimit = 136;
```
ycherniavskyi, line 12: Could you clarify the source of this constant value? If it originates from a specific place or reasoning, it would be helpful to include a comment with a link or at least a brief explanation.

Additionally, to avoid potential duplication, which we can already see within the file of this pull request, it would be beneficial to move this constant to `lib/app/constants.dart`.

lib/features/keypad/widgets/keypad.dart
```diff
@@ -1,24 +1,28 @@
...
-  final void Function(String) onKeypadPressed;
-
   @override
   Widget build(BuildContext context) {
     final themeData = Theme.of(context);
     final TextButtonStyles? textButtonStyles = themeData.extension<TextButtonStyles>();
 
     return TextButtonsTable(
-      minimumSize: Size.square(MediaQuery.of(context).size.width / 5),
+      minimumSize: Size.square(min(MediaQuery.of(context).size.width / 5, _maxSizeLimit)),
```
ycherniavskyi, line 25: This line seems to be doing quite a few operations. To improve readability and clarity, consider introducing a local final variable, perhaps named `dimension`. Assign all these calculations to this variable, then use `dimension` when calling the `Size.square` constructor.

ycherniavskyi (changes requested):
Please ensure to implement all applicable feedback from `keypad.dart` in the `actionpad.dart` file as well.

lib/features/keypad/widgets/keypad.dart
SERDUN: Done.

lib/features/keypad/widgets/keypad.dart
SERDUN: Done.

ycherniavskyi (approved):

PR #14, opened 2023-05-23
Сhange style resource

SERDUN:


packages/style/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
...
+environment:
+  sdk: '>=2.19.6 <3.0.0'
+
+dependencies:
+  flutter:
+    sdk: flutter
+
+  http: ^0.13.6
+  shared_preferences: ^2.0.18
+  cupertino_icons: ^1.0.2
```
ycherniavskyi, line 15: This is for sure an odd dependency for this package.

packages/style/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
+name: style
+description: Style synchronization module
+
+version: 1.0.0+1
+
+environment:
+  sdk: '>=2.19.6 <3.0.0'
```
ycherniavskyi, line 7: Update to the last Dart version with the whole branch rebase.

pubspec.yaml
```diff
@@ -59,6 +59,7 @@ dependencies:
   validators: ^3.0.0
   webview_flutter: ^4.0.6
   webview_flutter_web: ^0.2.2+1
+  yaml: ^3.1.1
```
ycherniavskyi, line 62: Do we really need to introduce the YAML format to the project just to handle the `publisher.yaml` file? Wouldn't it be simpler and more consistent to store the `publisher` data in a JSON format within a `publisher.json` file instead?

lib/app/view/app.dart
```diff
@@ -74,7 +77,7 @@ class _AppState extends State<App> {
                 localizationsDelegates: AppLocalizations.localizationsDelegates,
                 supportedLocales: AppLocalizations.supportedLocales,
                 restorationScopeId: 'App',
-                title: EnvironmentConfig.APP_NAME,
+                title: widget.themeSettings.appName ?? EnvironmentConfig.APP_NAME,
```
ycherniavskyi, line 80: If we're planning on storing the app name in the theme settings, I suggest we remove the `APP_NAME` environment config entirely. The same applies to the `APP_DESCRIPTION` environment config; let's move it to the theme settings instead.

lib/features/settings/features/about/view/about_screen.dart
```diff
@@ -38,8 +39,38 @@ class AboutScreen extends StatelessWidget {
...
+                              snapshot.data!,
+                              height: logoHeight,
+                            );
+                          } else {
+                            return SizedBox(
+                              height: logoHeight,
+                            );
+                          }
+                      }
+                    },
```
ycherniavskyi, line 73: Noticing similar widget structures in several areas is a clear indication that we should consider introducing a dedicated common widget to streamline this functionality.

lib/main.dart
```diff
@@ -13,6 +13,8 @@ import 'package:webtrit_phone/environment_config.dart';
 import 'package:webtrit_phone/pre_bootstrap/pre_bootstrap.dart';
 import 'package:webtrit_phone/repositories/repositories.dart';
 
+import 'package:style/style.dart';
```
ycherniavskyi, line 16: Incorrect import order.

lib/theme/theme_settings.dart
```diff
@@ -1,5 +1,11 @@
+import 'dart:typed_data';
+
 import 'package:flutter/material.dart';
 
+import 'package:flutter_svg/svg.dart';
+import 'package:rxdart/rxdart.dart';
+import 'package:webtrit_phone/app/assets.gen.dart';
```
ycherniavskyi, line 7: Missed newline before current import.

packages/style/lib/src/api/api.dart
```diff
@@ -0,0 +1,51 @@
+import 'dart:convert' as convert;
+
+import '../_http_client/_http_client.dart'
+    if (dart.library.html) '../_http_client/_http_client_html.dart'
+    if (dart.library.io) '../_http_client/_http_client_io.dart' as platform;
```
ycherniavskyi, line 5: It looks like with the release of [http 1.0.0](https://pub.dev/packages/http/versions/1.0.0), we might be able to eliminate the need for this custom multiplatform solution!

ycherniavskyi (changes requested):

packages/style/pubspec.yaml
SERDUN: Yes, it looks like cupertino_icons is redundant

pubspec.yaml
SERDUN: when using yaml, I was guided by the fact that this format is chosen by other plugins for generation, but in our case we don have code generation, we can use generation,I think it would be better to discuss

lib/app/view/app.dart
SERDUN: I agree, but this prompted me to think about how to set the name of the application in the system itself

lib/theme/theme_settings.dart
```diff
@@ -83,3 +93,64 @@ class ColorSchemeOverride {
   final Color? scrim;
   final Color? surfaceTint;
 }
+
+class ImagesScheme {
+  final onboardingStream = ReplaySubject<SvgLoader?>(maxSize: 1);
+  final applicationLogoStream = ReplaySubject<SvgLoader?>(maxSize: 1);
```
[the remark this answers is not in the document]
SERDUN: No, i don't know in dart streams about similar behaviour, ReplaySubject emits me last subject after connect to stream builder for example, even if it's happened before

lib/theme/theme_settings.dart
```diff
@@ -83,3 +93,64 @@ class ColorSchemeOverride {
...
+  }) {
+    _updateStreamByNetworkSvg(applicationLogoStream, image, svgGenImage);
+  }
+
+  void setApplicationByBytes(
+    Uint8List? image, {
+    SvgGenImage svgGenImage = Assets.logo,
+  }) {
+    _updateStreamByBytesSvg(applicationLogoStream, image, svgGenImage);
+  }
```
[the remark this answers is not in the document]
SERDUN: Because we have several use cases, for phone and for configuration. In configuration i triggered this method when user change resource

SERDUN:
The requests were divided into several smaller pools

PR #16, opened 2023-05-31
Migration of the bottom navigation according to the new approach in GoRouter

SERDUN:


lib/app/view/app.dart
```diff
@@ -216,176 +182,250 @@ class _AppState extends State<App> {
...
+                    final provider = BlocProvider(
+                      create: (context) {
+                        return MainBloc(
+                          infoRepository: context.read<InfoRepository>(),
+                          storeInfoExtractor: StoreInfoExtractor(),
+                        )..add(const MainStarted());
+                      },
+                      child: MainScreen(child),
+                    );
+
```
ycherniavskyi, line 200: Please remove the extra newline, as it's not consistent with the formatting in the rest of the similar code blocks.

ycherniavskyi (changes requested):

ycherniavskyi (approved):

PR #17, opened 2023-06-01
Remove spaces in phone property

SERDUN:
I checked FlutterContacts documentation, i don't see another way for delete "formatting" space.


ycherniavskyi (commented):
If I recall correctly, we agreed that this normalization should be handled by the app prior to the actual call, correct? If that's the case, we may need to close this PR without merging.

SERDUN:
> 

Hm, as i remember we agreed normalization it on ui, not in datasource. We can't close it because we will have bug which sent customer.

ycherniavskyi:
The actual removal of spaces in the phone number occurs prior to initiating the call - c22b7e05e2ddc4814dcbb84b7a7f3cf4b1b978b1.

PR #19, opened 2023-06-23
Revamp signup functionality to align with updated API

SERDUN:


packages/webtrit_api/lib/src/webtrit_api_client.dart
```diff
@@ -92,10 +99,10 @@ class WebtritApiClient {
     return Info.fromJson(responseJson);
   }
 
-  Future<SessionOtpProvisional> sessionOtpRequestDemo(SessionOtpCredentialDemo sessionOtpCredentialDemo) async {
-    final requestJson = sessionOtpCredentialDemo.toJson();
+  Future<SessionOtpProvisional> userSignupRequest(UserSignupCredentials sessionLoginCreateUser) async {
```
ycherniavskyi, line 102: Let's use the same method (and all other related names) base name as this action names in API - `createUser`.
image

ycherniavskyi (changes requested):
We should review and update the return values related to session and user in `WebtritApiClient`. This will ensure they are in sync with possible result schemas.

ycherniavskyi (approved):

PR #20, opened 2023-07-12
feat(webtrit_api): add x-request-id

SERDUN:


packages/webtrit_api/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
   json_annotation: ^4.8.1
   meta: ^1.9.1
   pub_semver: ^2.1.4
+  uuid: ^3.0.7
```
ycherniavskyi, line 14: I propose not to add the additional dependency for this functionality. `X-Request-ID` can be any random string. Let's implement it with Dart SDK functions. Here is how it is implemented in the Phoenix framework - [Plug.RequestId](https://github.com/elixir-plug/plug/blob/7b859ef85ede3154ff39249f0310064654d896e3/lib/plug/request_id.ex#L67).

packages/webtrit_api/lib/src/webtrit_api_client.dart
```diff
@@ -54,9 +55,12 @@ class WebtritApiClient {
       ],
     );
     final httpRequest = http.Request(method, url);
+    final requestId = Uuid().v4();
```
ycherniavskyi, line 58: Let's also introduce the optional `_httpClientExecute` function parameter `requestId`, and generate it our way if this parameter is `null`.

packages/webtrit_api/lib/src/webtrit_api_client.dart
```diff
@@ -54,9 +55,12 @@ class WebtritApiClient {
...
+
     httpRequest.headers.addAll({
       'content-type': 'application/json; charset=utf-8',
       'accept': 'application/json',
+      'X-request-id': requestId,
```
ycherniavskyi, line 63: Let's make `X-request-id` all low case, as other header keys, or `X-Request-ID` if you really want a mixed case way.

ycherniavskyi (changes requested):

ycherniavskyi (approved):

PR #25, opened 2023-07-27
feat: update texts resources on custom core url page

SERDUN:


lib/environment_config.dart
```diff
@@ -32,6 +32,16 @@ class EnvironmentConfig {
     defaultValue: 'WebTrit',
   );
 
+  static const VOIP_SYSTEM_NAME = String.fromEnvironment(
+    'WEBTRIT_VOIP_SYSTEM_NAME',
+    defaultValue: 'VoIP system',
+  );
```
ycherniavskyi, line 38: Hmm, I don't know if we need to add such a configuration. We will provide the possibility to customize all localization texts so that necessary information could be changed places directly there.

lib/environment_config.dart
```diff
@@ -32,6 +32,16 @@ class EnvironmentConfig {
...
+
+  static const SUPPORT_EMAIL = String.fromEnvironment(
+    'WEBTRIT_SUPPORT_EMAIL',
+    defaultValue: 'sales@portaone.com',
+  );
```
ycherniavskyi, line 43: First of all, there must be a `WEBTRIT_PHONE_` prefix, instead of just `WEBTRIT_`.
Then it is a bit odd for the variable with the name `SUPPORT_EMAIL` to have the default value `sales@portaone.com`.

So let's rename it to `SALES_EMAIL` and use `sales@webtrit.com` as the default value.

ycherniavskyi (changes requested):

lib/environment_config.dart
SERDUN: We don't have system now for change localization for different vendors, and look like it will be not in the nearest release. Maybe it has the right to live for some period of time until we implement localization

ycherniavskyi: In this case, I suggest rewording the sentence that requires this value. By doing so, we can make the text more generic and eliminate the need to specify the system name.

PR #26, opened 2023-07-31
Update phone with clarify Contact and expand UserInfo schemas

SERDUN:


packages/webtrit_api/test/webtrit_api_test.dart
```diff
@@ -526,9 +526,12 @@ void main() {
         return Response(
           jsonEncode({
             'sip': {
-              'display_name': 'display_name_1',
```
ycherniavskyi, line 529: This field didn't remove.

packages/webtrit_api/test/webtrit_api_test.dart
```diff
@@ -540,9 +543,9 @@ void main() {
               'ext': '0001',
               'additional': ['380441234567', '34911234567'],
             },
+            'email': 'email_1',
             'first_name': 'first_name_1',
             'last_name': 'last_name_1',
```
ycherniavskyi, line 548: Add `alies_name` field here.

ycherniavskyi (changes requested):
Within this PR also must be added DB migration for renaming `displayName` to `aliasName`.

ycherniavskyi (changes requested):
Also, it seems that you forgot to update the test data with values for added and renamed fields.

SERDUN:
It seems better not to use idea refactoring when replacing the name and do it by hand, for correctly style

ycherniavskyi (approved):

PR #27, opened 2023-08-02
Change contacts position for demo

SERDUN:
I did this approach in the development thread.
But in general, I don't like this solution, especially the fact that potentially the position of the contacts can be changed not only in the demo.

Is it possible to export the positioning logic to the configuration file? (Style)


ycherniavskyi:
Is this PR still relevant after the inclusion of #29, which sets the appropriate default tabs index value?

SERDUN:
> Is this PR still relevant after the inclusion of #29, which sets the appropriate default tabs index value?

Yes looks like not actual,I left the position of the tabs as they were, but by default the first tab will be opened immediately (or the last one in the list)

PR #28, opened 2023-08-02
feat: substitute warnings with errors for analysis.

SERDUN:


ycherniavskyi:
@SERDUN there is a lot of code duplication. Let's introduce our custom `flutter_lints` package (by analogic with [effective_dart](https://github.com/felangel/effective_dart)) and use it in all other packages. And then we could make such and other substitutions centralized.

PR #29, opened 2023-08-02
Feature to remember last tab position

SERDUN:


lib/data/app_preferences.dart
```diff
@@ -63,4 +63,13 @@ class AppPreferences {
...
   Future<bool> removeLocale() => _sharedPreferences.remove(_kLocaleLanguageTagKey);
+
+  int getActiveTabIndex(
+    String key, {
+    int defaultValue = 0,
+  }) {
+    return _sharedPreferences.getInt(key) ?? defaultValue;
+  }
+
+  Future<bool> setActiveTabIndex(String key, int index) => _sharedPreferences.setInt(key, index);
```
ycherniavskyi, line 74: Let's make this functions naming more general - `getActiveIndex` and `setActiveIndex`.

Also I propose to add guard prefix for passed `key`.
So instead of direct use of `key`, let's use `_activeIndexKey(key)`, where:
```dart
String _activeIndexKey(String indexKey) => 'index-key-${indexKey}';
```

lib/data/app_preferences.dart
```diff
@@ -63,4 +63,13 @@ class AppPreferences {
   Future<bool> setLocale(Locale value) => _sharedPreferences.setString(_kLocaleLanguageTagKey, value.toLanguageTag());
 
   Future<bool> removeLocale() => _sharedPreferences.remove(_kLocaleLanguageTagKey);
+
+  int getActiveTabIndex(
+    String key, {
```
ycherniavskyi, line 68: Let's make variable name more specific - `indexKey`.

ycherniavskyi (changes requested):
We need to find the way to generalise possibility for storing actual and restoring initial value for all `TabBar` widget (with help of `TabController`).

ycherniavskyi (approved):

PR #30, opened 2023-08-02
Change tab width

SERDUN:
Perhaps it is worth making a common constant


ycherniavskyi:
It seems that we can use special property for this task in future - https://github.com/flutter/flutter/commit/a732a7488807920dfd192b316584c620fa509ff7.

PR #31, opened 2023-08-03
Exception Flow Handling

SERDUN:


ycherniavskyi:
Implemented in a more straightforward manner within #39.

PR #36, opened 2023-08-08
Migrate to WebTrit App ids and configs

SERDUN:


SERDUN:
I see that the formatting in the manifest has changed

ycherniavskyi (changes requested):
@SERDUN, if you've decided to eliminate the functionality of the `WEBTRIT_PHONE_APP_NAME` and `WEBTRIT_PHONE_APP_ID_SUFFIX` environment variables, it should be done consistently across both Android and iOS platforms.

SERDUN:
I updated the ios as well, but I'm not sure if I did everything right on the ios side.

ycherniavskyi (approved):

PR #38, opened 2023-08-08
Feature/image provading

SERDUN:


lib/theme/theme_settings.dart
```diff
@@ -1,5 +1,12 @@
+import 'dart:typed_data';
+
 import 'package:flutter/material.dart';
 
+import 'package:flutter_svg/svg.dart';
+import 'package:rxdart/rxdart.dart';
```
ycherniavskyi, line 6: This dependency was removed, so try to re-implement its functionality another way.

lib/theme/theme_settings.dart
```diff
@@ -83,3 +92,60 @@ class ColorSchemeOverride {
...
+    SvgGenImage svgGenImage = Assets.logo,
+  }) {
+    if (url != null) {
+      return _secondaryOnboardingLogoStream.add(SvgNetworkLoader(url));
+    } else if (bytes != null) {
+      return _secondaryOnboardingLogoStream.add(SvgBytesLoader(bytes));
+    } else if (asset != null) {
+      return _secondaryOnboardingLogoStream.add(SvgAssetLoader(asset.path));
+    } else {
+      return _primaryOnboardingLogoStream.add(SvgAssetLoader(svgGenImage.path));
```
ycherniavskyi, line 140: Copy-past type.

ycherniavskyi (changes requested):

ycherniavskyi (changes requested):
To streamline the implementation, let's leverage the existing theme provider concept instead of adding a new SVG notifier. It might be beneficial to introduce a custom `SvgAsset` object, akin to `SvgGenImage`.

PR #40, opened 2023-08-23
Feature/color style

SERDUN:


lib/data/app_theme.dart
```diff
@@ -0,0 +1,68 @@
...
+  static Color? _toColor(String hex) {
+    try {
+      final buffer = StringBuffer();
+      if (hex.length == 6 || hex.length == 7) buffer.write('ff');
+      buffer.write(hex.replaceFirst('#', ''));
+      return Color(int.parse(buffer.toString(), radix: 16));
+    } catch (e) {
+      return null;
+    }
+  }
```
ycherniavskyi, line 67: I propose to make this logic as method `toColor` of `String` with extension.

lib/data/app_theme.dart
```diff
@@ -0,0 +1,68 @@
...
+    _instance = AppTheme._(theme);
+  }
+
+  factory AppTheme() => _instance;
+
+  AppTheme._(this._themeSettings);
+
+  ThemeSettings _themeSettings;
+
+  ThemeSettings get themeSettings => _themeSettings;
```
ycherniavskyi, line 23: `themeSettings` property name of `AppTheme` class is a bit redundant to me, so I propose to rename it to just `settings`. (`ThemeSettings` class name is ok).

assets/style/branding.json
ycherniavskyi, line 1: Let's use an upper case hex string of colors (usually graphical IDEs show hex color that way, so it will be easy to copy-past).

lib/data/app_theme.dart
```diff
@@ -0,0 +1,68 @@
...
+class AppTheme {
+  static late AppTheme _instance;
+
+  static Future<void> init() async {
+    final styleMap = jsonDecode(await rootBundle.loadString(Assets.style.branding));
+    ThemeSettings theme = _parseBrandingJson(styleMap);
+    _instance = AppTheme._(theme);
+  }
+
+  factory AppTheme() => _instance;
```
ycherniavskyi, line 17: Let's use the block body method format to be an analogy with other data singleton classes.

ycherniavskyi (changes requested):

ycherniavskyi (approved):

PR #43, opened 2023-09-06
Refactor/permission text

SERDUN:


lib/features/contacts/features/contacts_local_tab/view/contacts_local_tab.dart
```diff
@@ -24,15 +24,21 @@ class ContactsLocalTab extends StatelessWidget {
...
-            mainAxisAlignment: MainAxisAlignment.center,
-            children: [
-              Text(context.l10n.contacts_LocalTabText_permissionFailure),
-              TextButton(
-                onPressed: () => openAppSettings(),
-                child: Text(context.l10n.contacts_LocalTabButton_openAppSettings),
-              ),
-            ],
+          return Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 16),
```
ycherniavskyi, line 28: We already have a constant for such properties, named kAllPadding16. If you require a different value, please add it as a new constant and use that constant in the core code.

ycherniavskyi (changes requested):
Only a portion of the issue with unpadding text has been fixed. The 'else' case within the same method will also display text without padding, and the same issue occurs in the ContactsExternalTab widget.

It may be beneficial to introduce a generalized widget for handling this scenario, which can then be used in both ContactsLocalTab and ContactsExternalTab initially.

ycherniavskyi (approved):
The PR was extended for all no-data sections in the screens.

PR #44, opened 2023-09-11
Feature/webtrit theme

SERDUN:


ycherniavskyi:
After analyzing the resulting degradation API, because of moving from the `Color` type (which is not available in Dart SDK) to `String` one, it was decided to reject this refactoring.

PR #48, opened 2023-10-17
fix: center-align text within logo widget

SERDUN:


SERDUN:
When the welcome text is too large, it is positioned on the left, which does not look very good

ycherniavskyi (approved):

PR #51, opened 2023-10-27
Feature/login

SERDUN:


packages/webtrit_api/lib/src/models/system_info.dart
ycherniavskyi, line 1: Because the list of supported functionalities is free from values array, we can't introduce specific enum with all its values.
That is why let's introduce `CommonSupported` enum and add helper `commonSupported` getter which will return `List<CommonSupported>` to `AdapterInfo` class.

packages/webtrit_api/lib/src/models/system_info.dart
```diff
@@ -119,9 +119,21 @@ class AdapterInfo with _$AdapterInfo {
   const factory AdapterInfo({
     String? name,
     String? version,
-    List<String>? supported,
+    @Default([LoginType.otpSignin]) List<LoginType> supported,
```
ycherniavskyi, line 122: By default it must be emtry list.

ycherniavskyi (changes requested):

PR #52, opened 2023-11-02
Feature/launch pdf terms

SERDUN:


ios/Runner/Info.plist
```diff
@@ -67,5 +67,9 @@
 		<true/>
 		<key>UIStatusBarHidden</key>
 		<false/>
+		<key>LSApplicationQueriesSchemes~ipad</key>
```
ycherniavskyi, line 70: What is the reason to add `~ipad` to this key?

ios/Runner/Info.plist
```diff
@@ -67,5 +67,9 @@
...
+		<array>
+			<string>https</string>
```
ycherniavskyi, line 72: Let's also add `http` value.

android/app/src/main/AndroidManifest.xml
```diff
@@ -15,6 +15,13 @@
     <uses-permission android:name="android.permission.READ_CONTACTS"/>
     <uses-permission android:name="android.permission.WRITE_CONTACTS"/>
     <uses-permission android:name="android.permission.GET_ACCOUNTS"/>
+    <queries>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.BROWSABLE" />
+            <data android:scheme="https" />
+        </intent>
```
ycherniavskyi, line 23: Let's also add intent for `http` scheme.

lib/features/settings/view/settings_screen.dart
```diff
@@ -210,4 +217,10 @@ class SettingsScreen extends StatelessWidget {
       child: scaffold,
     );
   }
+
+  void _launchUrl(BuildContext context, Uri uri) async {
```
ycherniavskyi, line 221: The `context` parameter is unnecessary here.
And also let's give this method a more specific name - something like `_launchUrlWithDefaultPlatformViewer`.

lib/features/settings/view/settings_screen.dart
```diff
@@ -138,7 +139,13 @@ class SettingsScreen extends StatelessWidget {
               title: Text(context.l10n.settings_ListViewTileTitle_termsConditions),
               trailing: const Icon(Icons.navigate_next),
               onTap: () {
-                context.pushNamed(MainRoute.settingsTermsConditions);
+                final uri = Uri.parse(EnvironmentConfig.APP_TERMS_AND_CONDITIONS_URL);
+
+                if (uri.path.endsWith('.pdf')) {
+                  _launchUrl(context, uri);
+                } else {
+                  context.pushNamed(MainRoute.settingsTermsConditions);
```
ycherniavskyi, line 147: It seems that we could make some refactoring with this PR, and pass the "terms and conditions URL" as a routing parameter to the `initialUri` property of the `TermsConditionsScreen` widget.

ycherniavskyi (changes requested):

ios/Runner/Info.plist
SERDUN: thank you, I can't understand the world it took, I'll delete it

lib/features/settings/view/settings_screen.dart
SERDUN: okay ,  good point.
What about key ''initialUrl', should we move it in const ? but i not sure where it better to keep  consts.dart or routes.dart

SERDUN: tried to bring out the hard coded values in route queries keys

ycherniavskyi (approved):

PR #54, opened 2023-11-07
refactor(Android): migrate from using key.property to key.json

SERDUN:


android/app/build.gradle
```diff
@@ -27,11 +29,8 @@ def dartDefine = [
     WEBTRIT_APP_NAME: project.hasProperty('WEBTRIT_APP_NAME') ? WEBTRIT_APP_NAME : 'WebTrit',
 ]
 
-def keystoreProperties = new Properties()
-def keystorePropertiesFile = rootProject.file('key.properties')
-if (keystorePropertiesFile.exists()) {
-    keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
-}
+def keystoreJsonFile = file('../key.json')
```
ycherniavskyi, line 32: I propose to introduce the optional `WEBTRIT_ANDROID_RELEASE_KEY_JSON_PATH` environment variable to pass some external `key.json` file path.

android/app/build.gradle
```diff
@@ -64,10 +63,10 @@ android {
 
     signingConfigs {
         release {
-            keyAlias keystoreProperties['keyAlias']
-            keyPassword keystoreProperties['keyPassword']
-            storeFile keystoreProperties['storeFile'] ? file(keystoreProperties['storeFile']) : null
-            storePassword keystoreProperties['storePassword']
+            keyAlias keystoreJson.alias
```
ycherniavskyi, line 66: Let's use the same naming as in the release section, so instead of `alias` must be `keyAlias`.

ycherniavskyi (changes requested):

android/app/build.gradle
SERDUN: I propose to make it even more flexible in this case, and transfer the path not to json but to the package where key.json and upload-keystore.jks are located, it will work well with our key repository, and more flexibly define the path to the keys

ycherniavskyi:
Reimplemented within ddcc0ef2de613ff76e5320653ce2f0339deca36b.

PR #57, opened 2023-11-23
Feature/localizely

SERDUN:


ycherniavskyi:
Current PR adds overabundant unready locales and also utilizes excess dependencies. That is why it is closed in favor of #60.

PR #60, opened 2023-11-25
Localizely integration

ycherniavskyi:
Perform [Localizely](https://localizely.com) integration with addition uk locale as an example.


ycherniavskyi:
More details about Localizely CLI:
- https://localizely.com/cli/
- https://localizely.com/configuration-file/

PR #69, opened 2024-03-20
fix: clean keypad on call start or transfer

digiboridev:


ycherniavskyi (changes requested):
The current implementation is technically correct.
However, it introduces duplication of the same behavior - get and then clear. It also obscures the intention to have the same behavior for both call and transfer actions.
I propose introducing a `String _popNumber()` method in `KeypadViewState` to consolidate this functionality and clarify the intended behavior for both actions.

digiboridev:
ok

ycherniavskyi (approved):

PR #72, opened 2024-03-22
fix: autoroute.of missing, zone scope, test setup

digiboridev:


screenshots/android/app/build.gradle
```diff
@@ -44,7 +44,7 @@ android {
 
     defaultConfig {
         applicationId "com.webtrit.phone.screenshots"
-        minSdkVersion 23 // flutter.minSdkVersion
+        minSdkVersion 26 // flutter.minSdkVersion
```
ycherniavskyi, line 47: Such modification must be provided at least in separate commit with brief explanation of the reason of it. But even better in separate Pull request, because it is not directly related to current PR.

screenshots/integration_test/take_screenshots_test.dart
```diff
@@ -18,15 +18,34 @@ import 'package:webtrit_phone/features/features.dart';
...
+void main() {
   final binding = IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
-  await AppThemes.init();
-  final themeSettings = AppThemes().values.first.settings;
-
   late final String screenshotNamePrefix;
   late final AppBloc appBloc;
 
+  setUpAll(() async {
```
ycherniavskyi, line 27: What is the reason to move this method up? Previously it located before test methods but after helper method `takeScreenshotTestWidgets` - what was wrong with it?

screenshots/lib/widgets/screenshot_app.dart
```diff
@@ -1,12 +1,16 @@
...
 
 import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:mocktail/mocktail.dart';
 
 import 'package:webtrit_phone/blocs/app/app_bloc.dart';
 import 'package:webtrit_phone/environment_config.dart';
 import 'package:webtrit_phone/l10n/l10n.dart';
 import 'package:webtrit_phone/theme/theme.dart';
 
+import 'package:auto_route/src/router/controller/pageless_routes_observer.dart';
```
ycherniavskyi, line 12: Usually `import` from `src` is undesirable. Do we really need this direct `import`?

ycherniavskyi (changes requested):

screenshots/android/app/build.gradle
digiboridev: Callkeep requires minsdk26, need extra time to understand why, i support the idea of separate commit.

screenshots/integration_test/take_screenshots_test.dart
digiboridev: its not about position its about do all async work inside setUpAll function before tests. Driver not support async work inside test functions

screenshots/lib/widgets/screenshot_app.dart
digiboridev: Yes its bad side of that tipe of fix. I dont know why auto_route hasnt ability to mock its router stack. I think better fix for that its completely get rid of router(low level deps i think) inside shared code, and make dumb reconstruction of screens using decomposition.

digiboridev:
discard

PR #73, opened 2024-03-24
feat: add user agreement checkbox to permissions screen

SERDUN:
This is a requirement of reviewers from the play store, they do not allow updating applications without an explicit reference to the terms and conditions when requesting permissions.

Reviewers note:
_Issue with your app:

Your app is uploading users' Contact List information but the prominent disclosure does not meet the policy requirements.

Your app does not have a prominent disclosure.
Action Required:

Add a valid prominent disclosure within your app. Please review Best practices for prominent disclosure as guidelines for your prominent disclosure._


lib/app/router/app_router.dart
ycherniavskyi, line 1: I propose to create separate screen for permissions terms and conditions - `PermissionsTermsConditionsScreen`. Because technicaly they have difference intend: `TermsConditionsScreen` just show information but `PermissionsTermsConditionsScreen` can show information and provide action about it.

lib/data/app_preferences.dart
```diff
@@ -13,6 +13,7 @@ class AppPreferences {
   static const _kActiveMainFlavorKey = 'active-main-flavor';
   static const _kActiveRecentsVisibilityFilterKey = 'active-recents-visibility-filter';
   static const _kActiveContactSourceTypeKey = 'active-contact-source-type';
+  static const _kUserAgreement = 'user-agreement';
```
ycherniavskyi, line 16: Let's use more solid naming here (as actually was done in following code) - 'user-agreement-accepted' (and for all related variables and methods corresponding - `setUserAgreementAccepted`, `setUserAgreementAccepted`, etc.).

lib/features/permissions/cubit/permissions_cubit.dart
ycherniavskyi, line 1: What about instead of complicate `PermissionsCubit` and `PermissionsState` by introducing `appPreferences` property and its state, just extend a bit `AppBloc` and `AppState` with 'user-agreement-accepted' parameter and just use `AppSate` in following widgets.

lib/features/permissions/widgets/agreement_checkbox.dart
```diff
@@ -0,0 +1,57 @@
...
+                  ),
+                  recognizer: TapGestureRecognizer()..onTap = onAgreementLinkTap,
+                ),
+                TextSpan(
+                  text: context.l10n.permission_agreement_text3,
+                  style: themeData.textTheme.labelLarge,
+                ),
+              ],
+            ),
+          ),
```
ycherniavskyi, line 52: Let's try to refactor this part with [parameterized l10n](https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization#placeholders-plurals-and-selects) and `Linkify` widget within the project (that could be extended if necessary with possibility to open internal page).

ycherniavskyi (changes requested):

PR #81, opened 2024-04-01
fix (screenshots_app): integration test auto_route context fix candidate

digiboridev:
Fix candidate for auto_route router stack can't finded in context durning widget test

### DIFF
--- a/screenshots/lib/widgets/screenshot_app.dart
+++ b/screenshots/lib/widgets/screenshot_app.dart
@@ -1,11 +1,16 @@
+import 'package:auto_route/auto_route.dart';
 import 'package:flutter/material.dart';
 
 import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:mocktail/mocktail.dart';
 
 import 'package:webtrit_phone/blocs/app/app_bloc.dart';
 import 'package:webtrit_phone/environment_config.dart';
 import 'package:webtrit_phone/l10n/l10n.dart';
 import 'package:webtrit_phone/theme/theme.dart';
+
+// Attention: undesired import, but required for the mock auto_route router stack
+import 'package:auto_route/src/router/controller/pageless_routes_observer.dart';
 
 class ScreenshotApp extends StatelessWidget {
   const ScreenshotApp({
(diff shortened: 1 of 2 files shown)


screenshots/lib/widgets/screenshot_app.dart
```diff
@@ -1,12 +1,17 @@
+import 'package:auto_route/auto_route.dart';
```
SERDUN, line 1: Please group imports according to our style.
https://app.clickup.com/4775009/docs/4hq31-181/4hq31-1068

screenshots/lib/widgets/screenshot_app.dart
```diff
@@ -1,12 +1,17 @@
...
 import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:mocktail/mocktail.dart';
 
 import 'package:webtrit_phone/blocs/app/app_bloc.dart';
 import 'package:webtrit_phone/environment_config.dart';
 import 'package:webtrit_phone/l10n/l10n.dart';
 import 'package:webtrit_phone/theme/theme.dart';
 
+// Attention: undesired import, but required for the mock auto_route router stack
+import 'package:auto_route/src/router/controller/pageless_routes_observer.dart';
```
SERDUN, line 13: Please group imports according to our style.
https://app.clickup.com/4775009/docs/4hq31-181/4hq31-1068

I think it is worth leaving a link to the issue in github for easy status tracking

screenshots/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
   clock: ^1.1.1
   device_info_plus: ^9.1.0
   flutter_bloc: ^8.1.3
+  auto_route: ^7.8.4
```
SERDUN, line 19: You are adding a dependency that we use for navigation, but I don't see anywhere that the previous navigation plugin was removed.

Perhaps this is in a separate commit altogether, as I also don't see the removal of the previous navigation logic in other branches.
https://github.com/WebTrit/webtrit_phone/blob/fix/screenshots_integration_test_autoroute_context/screenshots/lib/main.dart


I think we should stick to using one navigation plugin in sub-modules according to the main project.

PR #128, opened 2024-04-25
Fix (autoprovision): token exchange before confirmation

digiboridev:
Attention dont accept before fix "401 session missing" from backend side!!


digiboridev:
Droped due flow change discard

PR #130, opened 2024-04-25
Feat: attended transfer impl

digiboridev:


lib/features/call/bloc/call_event.dart
```diff
@@ -195,6 +195,9 @@ class CallControlEvent with _$CallControlEvent implements CallEvent {
   const factory CallControlEvent.blindTransferInitiated(String callId) = _CallControlEventBlindTransferInitiated;
 
   const factory CallControlEvent.blindTransferred({required String number}) = _CallControlEventBlindTransferred;
+
+  const factory CallControlEvent.attendedTransferred(
+      {required ActiveCall referorCall, required ActiveCall transfereeCall}) = _CallControlEventAttendedTransferred;
```
SERDUN, line 200: If we have more than one named parameter, let's use comma+ formatting.

  const factory CallControlEvent.attendedTransferred({
    required ActiveCall referorCall,
    required ActiveCall transfereeCall,
  }) = _CallControlEventAttendedTransferred;

lib/features/call/bloc/call_state.dart
```diff
@@ -104,9 +104,9 @@ class CallState with _$CallState {
     return copyWith(activeCalls: activeCalls);
   }
 
-  CallState copyWithPushActiveCall(ActiveCall activeCall) {
+  CallState copyWithPushActiveCall(ActiveCall activeCall, {bool? minimized}) {
```
ycherniavskyi, line 107: `minimized` is surplus here.
If you need to change it after this function execution then just use following construction:
```dart
emit(state.copyWithPushActiveCall(ActiveCall(<properties>)).copyWith(minimized: <value>));
```

lib/features/call/bloc/call_state.dart
```diff
@@ -152,6 +152,7 @@ class ActiveCall with _$ActiveCall {
 
 extension ActiveCallIterableExtension<T extends ActiveCall> on Iterable<T> {
   T get current => lastWhere((activeCall) => !activeCall.held, orElse: () => last);
+  List<T> get exceptCurrent => where((activeCall) => activeCall != current).toList();
```
ycherniavskyi, line 155: `exceptCurrent` name is hide actual state of this calls - they are `held`, so let's rename this getter to `held`.
And in implementation let's filter by `activeCall.held` instead of by `current` equality.

lib/features/call/view/call_active_scaffold.dart
```diff
@@ -39,6 +39,7 @@ class CallActiveScaffoldState extends State<CallActiveScaffold> {
   Widget build(BuildContext context) {
     final activeCalls = widget.activeCalls;
     final activeCall = activeCalls.current;
+    final backgroundCalls = activeCalls.exceptCurrent;
```
ycherniavskyi, line 42: `backgroundCalls` let's rename to `heldCalls`.

lib/features/call/widgets/call_actions.dart
```diff
@@ -131,9 +136,11 @@ class _CallActionsState extends State<CallActions> {
     final onMutedChanged = widget.onMutedChanged;
     final speakerValue = widget.speakerValue;
     final onSpeakerChanged = widget.onSpeakerChanged;
-    final onTransferPressed = widget.onTransferPressed;
+    final onBlindTransfer = widget.onBlindTransfer;
+    final onAttendedTransfer = widget.onAttendedTransfer;
     final onHeldChanged = widget.onHeldChanged;
     final onSwapPressed = widget.onSwapPressed;
+    final themeData = Theme.of(context);
```
ycherniavskyi, line 143: Usually we put general variable such as `themeData` define before more local ones and separate them by one new line. So please move it before this handle variables block.

ycherniavskyi (changes requested):
It seems that some commits descriptions are not following [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) style, please do not forget about it.

lib/features/call/bloc/call_state.dart
digiboridev: @ycherniavskyi there is reason, because of "orElse: () => last" in current getter. So this list is literaly "except current" not "held"

lib/features/call/bloc/call_bloc.dart
```diff
@@ -1166,9 +1172,29 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
...
+      final transferRequest = TransferRequest(
+        transaction: WebtritSignalingClient.generateTransactionId(),
+        line: referorCall.line,
+        callId: referorCall.callId.toString(),
+        number: referorCall.handle.normalizedValue(),
+        replaceCallId: transfereeCall.callId,
+      );
 
-      _peerConnectionCompleteError(activeCallBlindTransferInitiated.callId, e);
-      add(_ResetStateEvent.completeCall(activeCallBlindTransferInitiated.callId));
```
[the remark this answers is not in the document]
digiboridev: @ycherniavskyi this lines i added few days ago, for now i have deeper understanding of flow and rethink its purpose.  I think there is no need to drop active call if transfer request fails, it can be caused by adapter as unrecognized for example, so nothing stops user to just continue active conversation if its possible.

lib/features/call/bloc/call_state.dart
ycherniavskyi: Clear. In such case I propose to name this getter `notCurrent`.

PR #138, opened 2024-05-14
feat: build version env

digiboridev:


SERDUN:
I planned to simply update the pubspec.yaml (version: ..+...) in the configure phase, and then the old method will work (reducing the amount of boilerplate).
Let the pull request will in draft status some time.

SERDUN:
It looks like we can simplify it by using: flutter build --build-name=1.2.0+1 --build-number=2

SERDUN:
Will use cli for set build name and number

PR #164, opened 2024-06-18
feat: migrate to the new theme scheme

SERDUN:


SERDUN:
Don't ready, should be updated scheme on server side

ycherniavskyi (approved):

PR #168, opened 2024-06-25
Build proper tenant URL

ycherniavskyi:


ycherniavskyi:
This PR implements the functionality of #166 and #167, albeit in a slightly simpler way. A drawback of the current implementation is the duplication of the `buildTenantUrl` methods, but this can be refactored in the future.

PR #171, opened 2024-06-28
Integration tests integration with login cases

digiboridev:


dart_define.json
ycherniavskyi, line 1: Let's create separate file `dart_define_test.json` with test defines.
`--dart-define-from-file` can be used multiple times.

ycherniavskyi (changes requested):

PR #173, opened 2024-06-28
fix: sqlite crash loop on failed migration

digiboridev:


packages/data/app_database/lib/src/app_database.dart
```diff
@@ -44,9 +46,29 @@ class AppDatabase extends _$AppDatabase {
   MigrationStrategy get migration {
     return MigrationStrategy(
       onUpgrade: (m, from, to) async {
-        for (var version = from; version < to; version++) {
-          await migrations[version - 1].execute(this, m);
-        }
+        await customStatement('PRAGMA foreign_keys = OFF');
```
ycherniavskyi, line 49: What is the reason for disabling `foreign_keys` for all migrations? While I understand that some migrations may require this, it might be better to disable `foreign_keys` check only when necessary during execution of particular migration.

packages/data/app_database/lib/src/app_database.dart
```diff
@@ -44,9 +46,29 @@ class AppDatabase extends _$AppDatabase {
...
+
+        // Use a transaction to avoid concurrent migration
+        // with another isolate (e.g. FCM isolate init [bootstrap.dart - _firebaseMessagingBackgroundHandler])
+        await transaction(() async {
```
ycherniavskyi, line 53: This behavior is unusual because previously Drift was configured to run in [“isolate mode”](https://drift.simonbinder.eu/docs/advanced-features/isolates/), where other isolates do not directly access the database file but do so through the Drift isolate. It seems necessary to verify if Drift is correctly configured for this mode of operation, as in such cases issues like concurrent migration should not occur.

packages/data/app_database/lib/src/app_database.dart
```diff
@@ -44,9 +46,29 @@ class AppDatabase extends _$AppDatabase {
...
+            if (remoteCause is SqliteException && remoteCause.resultCode == 1) {
+              // If a migration fails, drop all entities and recreate them in actual form
+              // At least to provide app functionality with clean data instead of crashing loop
+              for (final e in allSchemaEntities) {
+                await m.drop(e);
+              }
+              await m.createAll();
+            } else {
+              rethrow;
+            }
```
ycherniavskyi, line 69: This behavior is certainly undesirable, but handling it silently is even worse. Let’s explore how we can notify the user about the inability to apply the migration and suggest clearing the database.

ycherniavskyi (changes requested):

PR #175, opened 2024-07-02
fix: wrong ids for removed contacts while sync

digiboridev:


ycherniavskyi (changes requested):
It seems that `getIdsBySourceType` method could be removed as redundant.

PR #177, opened 2024-07-04
feat: request credentials feature

SERDUN:


lib/app/router/app_router.dart
```diff
@@ -58,6 +58,10 @@ class AppRouter extends _$AppRouter {
                 AutoRoute(
                   page: LoginModeSelectScreenPageRoute.page,
                 ),
+                AutoRoute(
+                  page: LoginRequestCredentialsScreenPageRoute.page,
```
ycherniavskyi, line 62: Let's use one sequence of `credentials` and `request` words.
If env is `WEBTRIT_APP_CREDENTIALS_REQUEST_URL` then it will be more logical to use `LoginCredentialsRequestScreenPageRoute` instead of `LoginRequestCredentialsScreenPageRoute`.

lib/features/login/view/login_mode_select_screen.dart
```diff
@@ -35,6 +35,7 @@ class LoginModeSelectScreen extends StatelessWidget {
       buildWhen: (previous, current) => previous.processing != current.processing,
       builder: (context, state) {
         final isDemoModeEnabled = context.read<LoginCubit>().isDemoModeEnabled;
+        final isSignUpRequestEnabled = context.read<LoginCubit>().isCredentialRequestUrlEnabled;
```
ycherniavskyi, line 38: What is the reason to have different variable and method names?

lib/features/login/view/login_request_credentials_screen.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:flutter/material.dart';
+
+import 'package:webtrit_phone/features/settings/widgets/widgets.dart';
```
ycherniavskyi, line 3: Just import all features `package:webtrit_phone/features/features.dart`

lib/features/login/view/login_request_credentials_screen.dart
```diff
@@ -0,0 +1,35 @@
...
+import 'package:webtrit_phone/l10n/l10n.dart';
+import 'package:webtrit_phone/widgets/widgets.dart';
+
+class LoginRequestCredentialsScreen extends StatelessWidget {
+  const LoginRequestCredentialsScreen({
+    super.key,
+    required this.initialUri,
+  });
+
+  final Uri initialUri;
```
ycherniavskyi, line 13: Within our code base we usually use `url` term for variable name, so let's name this and all other added variable with `url` suffix.

ycherniavskyi (changes requested):

lib/features/login/view/login_mode_select_screen.dart
SERDUN: will fix this, forgot to change after changing the name of the feature.

PR #196, opened 2024-07-23
feat: attended transfer handle invite call

SERDUN:


lib/features/call/bloc/call_bloc.dart
```diff
@@ -720,6 +720,7 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
         displayName: event.callerDisplayName,
         video: video,
         createdTime: clock.now(),
+        transfer: Transfer.tryCreateAttemptInviteTransfer(event.referredBy, event.replaceCallId),
```
ycherniavskyi, line 723: As for me `tryCreateAttemptInviteTransfer` method is redundant here. Just init local variable `transfer` with proper value before `emit` method call.

lib/features/call/view/call_active_scaffold.dart
```diff
@@ -173,6 +173,7 @@ class CallActiveScaffoldState extends State<CallActiveScaffold> {
                       wasAccepted: activeCall.wasAccepted,
                       wasHungUp: activeCall.wasHungUp,
                       cameraValue: cameraEnabled,
+                      transferAttemptInvite: activeCall.transfer?.attemptInviteTransfer ?? false,
```
ycherniavskyi, line 176: First of all it is a bit odd that this properties have different names. And as for name, I propose to change it to `inviteToAttendedTransfer`.

ycherniavskyi (changes requested):

PR #204, opened 2024-07-31
feat: extend database contacts DTO to support generating thumbnail URL via Gravatar API

SERDUN:


ycherniavskyi:
Task linked: [CU-86958uxfp Add correct flow with Gravatar](https://app.clickup.com/t/86958uxfp)

lib/features/contact/view/contact_screen.dart
```diff
@@ -37,6 +37,7 @@ class ContactScreen extends StatelessWidget {
                     Padding(
                       padding: kAllPadding16,
                       child: LeadingAvatar(
+                        thumbnailUrl: contact.thumbnailUrl,
```
ycherniavskyi, line 40: Let's maintain a consistent order for local parameters. In this case, since `thumbnailUrl` is declared after `thumbnail`, let's use the same order during usage.

ycherniavskyi: Apply this logic on whole code in this PR, please.

lib/features/settings/widgets/user_info_list_tile.dart
```diff
@@ -48,6 +50,9 @@ class UserInfoListTile extends StatelessWidget {
           children: [
             LeadingAvatar(
               username: info?.name ?? '?',
+              thumbnailUrl: info?.email != null
+                  ? Gravatar(info!.email!).image(scheme: 'https', defaultImage: DefaultImage.fileNotFound).toString()
+                  : null,
```
ycherniavskyi, line 55: It seems we need to introduce `lib/utils/gravatar.dart` file with function `Uri? gravatarThumbnailUrl(String? email)`.

lib/widgets/leading_avatar.dart
```diff
@@ -17,44 +18,26 @@ class LeadingAvatar extends StatelessWidget {
 
   final String? username;
   final Uint8List? thumbnail;
+  final String? thumbnailUrl;
```
ycherniavskyi, line 21: It is a bit odd that `thumbnailUrl` is passed as `String?` type instead of `Uri?`. Yes, I see that `NetworkImage` requires `url` parameter in `String`, but let's just cast `Uri` to `String` just before passing it to `NetworkImage` constructor.

lib/repositories/contacts/contacts_repository.dart
```diff
@@ -52,16 +54,22 @@ class ContactsRepository {
     return _appDatabase.favoritesDao.deleteByContactPhoneId(contactPhone.id);
   }
 
-  Contact _toContact(ContactData data) {
+  Contact _toContact(ContactWithEmailData data) {
+    final email = data.email?.address;
+    final gravatarUrl = email != null
+        ? Gravatar(email).image(scheme: 'https', defaultImage: DefaultImage.fileNotFound).toString()
+        : null;
```
ycherniavskyi, line 61: Here we need to use previously declared `gravatarThumbnailUrl` function.

packages/data/app_database/lib/src/app_database.dart
ycherniavskyi, line 1: It seems that current change introduce some half backed functionality.
It does not populate returned composed contacts with all email, but only with first one, and also it does not add phones at all.
I cloud propose to left original method in `ContactsDao` as it was before, but add several `*ContactsExt` methods that will return following joining class:
```dart
class ContactWithPhonesAndEmailsData {
  ContactWithEmailData({
    required this.contact,
    this.email,
  });

  final ContactData contact;
  final List<ContactPhoneData> phones;
  final List<ContactEmailData> emails;
}
```

ycherniavskyi (changes requested):

packages/data/app_database/lib/src/app_database.dart
SERDUN: But in this case, the original methods will not be used, because the current widgets will need an email

lib/features/contacts/features/contacts_external_tab/view/contacts_external_tab.dart
```diff
@@ -30,6 +30,7 @@ class ContactsExternalTab extends StatelessWidget {
               return ContactTile(
                 displayName: contact.name,
                 registered: contact.registered,
+                thumbnailUrl: contact.thumbnailUrl,
```
ycherniavskyi, line 33: It seems that before this line we also need to add `thumbnail: contact.thumbnail,` to be consistent with `view/contacts_local_tab.dart`, even if we know that local contacts do not have binary thumbnail for now.

lib/features/contacts/widgets/contact_tile.dart
```diff
@@ -31,6 +33,7 @@ class ContactTile extends StatelessWidget {
         thumbnail: thumbnail,
         registered: registered,
         smart: smart,
+        thumbnailUrl: thumbnailUrl,
```
ycherniavskyi, line 36: Please, put this line after `thumbnail: thumbnail,`, so it will be in the same order as in widget declaration.

packages/data/app_database/lib/src/app_database.dart
```diff
@@ -315,6 +316,59 @@ class ContactsDao extends DatabaseAccessor<AppDatabase> with _$ContactsDaoMixin
     return (select(contactsTable)..whereSamePrimaryKey(contact)).watchSingle();
   }
 
+  Stream<List<ContactWithPhonesAndEmailsData>> watchAllContactsWithPhonesAndEmailsExt([
```
ycherniavskyi, line 319: To stay consistent with the naming conventions of other Dao class methods, this method should be named `watchAllContactsExt`. (The purpose of the `ext` suffix can be inferred from the returned class name.)

ycherniavskyi (changes requested):

ycherniavskyi (approved):

PR #205, opened 2024-07-31
feat: replace sqlite "like" command with "regexp" for utf-8 support

SERDUN:
https://www.sqlite.org/faq.html#q18

The default configuration of SQLite only supports case-insensitive comparisons of ASCII characters. The reason for this is that doing full Unicode case-insensitive comparisons and case conversions requires tables and logic that would nearly double the size of the SQLite library. The SQLite developers reason that any application that needs full Unicode case support probably already has the necessary tables and functions and so SQLite should not take up space to duplicate this ability.
Instead of providing full Unicode case support by default, SQLite provides the ability to link against external Unicode comparison and conversion routines. The application can overload the built-in [NOCASE](https://www.sqlite.org/datatype3.html#collation) collating sequence (using [sqlite3_create_collation()](https://www.sqlite.org/c3ref/create_collation.html)) and the built-in [like()](https://www.sqlite.org/lang_corefunc.html#like), [upper()](https://www.sqlite.org/lang_corefunc.html#upper), and [lower()](https://www.sqlite.org/lang_corefunc.html#lower) functions (using [sqlite3_create_function()](https://www.sqlite.org/c3ref/create_function.html)). The SQLite source code includes an "ICU" extension that does these overloads. Or, developers can write their own overloads based on their own Unicode-aware comparison routines already contained within their project.

### DIFF
--- a/packages/data/app_database/lib/src/app_database.dart
+++ b/packages/data/app_database/lib/src/app_database.dart
@@ -304,7 +304,7 @@ class ContactsDao extends DatabaseAccessor<AppDatabase> with _$ContactsDaoMixin 
           contactsTable.firstName,
           contactsTable.aliasName,
           contactPhonesTable.number,
-        ].map((c) => c.like('%$searchBit%')).reduce((v, e) => v | e),
+        ].map((c) => c.regexp('.*$searchBit.*', caseSensitive: false)).reduce((v, e) => v | e),
       );
     }
     q.groupBy([contactPhonesTable.contactId]);
--- a/packages/data/app_database/test/contacts_dao_test.dart
+++ b/packages/data/app_database/test/contacts_dao_test.dart
@@ -1,0 +1,100 @@
+import 'package:drift/native.dart';
+import 'package:test/test.dart';
+
+import 'package:app_database/app_database.dart';
+
+void main() {
+  late AppDatabase database;
+
+  setUp(() async {
+    database = AppDatabase(NativeDatabase.memory());
+
+    final contactsDao = database.contactsDao;
+
+    await contactsDao.insertOnUniqueConflictUpdateContact(ContactDataCompanion(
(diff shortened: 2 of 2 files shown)


ycherniavskyi:
Task linked: [CU-86954cwkw Contact search improvement](https://app.clickup.com/t/86954cwkw)

packages/data/app_database/lib/src/app_database.dart
```diff
@@ -295,18 +295,24 @@ class ContactsDao extends DatabaseAccessor<AppDatabase> with _$ContactsDaoMixin
...
       leftOuterJoin(contactPhonesTable, contactPhonesTable.contactId.equalsExp(contactsTable.id)),
     ]);
-    for (final searchBit in searchBits) {
-      q.where(
-        [
-          contactsTable.lastName,
-          contactsTable.firstName,
-          contactsTable.aliasName,
-          contactPhonesTable.number,
-        ].map((c) => c.like('%$searchBit%')).reduce((v, e) => v | e),
```
ycherniavskyi, line 305: Hmm, It seems that the same functionality could be achieved by updating columns map function to `c.lower().like('%${searchBit.toLowerCase()}%')`. If yes then I differently prefer such change that what PR provided currency.

ycherniavskyi (changes requested):

packages/data/app_database/lib/src/app_database.dart
SERDUN: Here's the corrected translation into English:

---

Yesterday, I tried this option, but it didn't help, although logically it should have, given the SQLite queries. We need to check the implementation of the LIKE function; perhaps it works with sequences after all. I agree that the current implementation is somewhat cumbersome. Therefore, I suggest we start by covering this function with tests to simplify debugging, and then continue our investigation.

ycherniavskyi (approved):

PR #206, opened 2024-08-01
feat: add handler for self-care password expired

SERDUN:


ycherniavskyi:
Task linked: [CU-86958met0 Application does not show the real error message](https://app.clickup.com/t/86958met0)

lib/blocs/app/app_bloc.dart
```diff
@@ -17,6 +19,17 @@ part 'app_event.dart';
 
 part 'app_state.dart';
 
+WebtritApiClient defaultCreateWebtritApiClient(String coreUrl, String tenantId) {
```
SERDUN, line 22: This code is duplicated from the "login" flow. It might make sense to move it to a common place for better code organization and reuse.

lib/blocs/app/app_bloc.dart
```diff
@@ -84,6 +108,10 @@ class AppBloc extends Bloc<AppEvent, AppState> {
     ));
   }
 
+  void _onLogoutedTeardown(AppLogoutTeardown event, Emitter<AppState> emit) async {
```
ycherniavskyi, line 111: `_onLogoutedTeardown` and `AppLogoutTeardown` why in one name `Logouted` and in other `Logout`?

ycherniavskyi (changes requested):

SERDUN:
Another point related to this pull request, for my demo task I need an email, which I often have to retrieve through the getUser method. Maybe we should add the user to the application state to avoid making the same requests frequently.

For example:
`  Future<DemoData> _getAction(String tab) async {
    final account = await _webtritApiClient.getUserInfo(_token);
    _logger.info('_prepareRequest: $account');

    return _webtritApiClient.createDemoData(DemoCredential(
      type: _platformInfo.appType,
      identifier: _appInfo.identifier,
      email: account.email!,
      tenantId: _tenantId,
      action: tab,
    ));
  }`

lib/blocs/app/app_bloc.dart
SERDUN: my mistake, thanks, I'll fix it

SERDUN:
> Another point related to this pull request, for my demo task I need an email, which I often have to retrieve through the getUserInfo method. Maybe we should add the user to the application state to avoid making the same requests frequently.
> 
> For example:
> 
> ```
>   Future<DemoData> _getAction(String tab) async {
>     final account = await _webtritApiClient.getUserInfo(_token);
>     _logger.info('_prepareRequest: $account');
> 
>     return _webtritApiClient.createDemoData(DemoCredential(
>       type: _platformInfo.appType,
>       identifier: _appInfo.identifier,
>       email: account.email!,
>       tenantId: _tenantId,
>       action: tab,
>     ));
>   }
> ```

or maybe just pull email from login flow   void _onLogined(AppLogined event, Emitter<AppState> emit) async {
____
@Freezed(copyWith: false)
class AppLogined with _$AppLogined implements AppEvent {
  const factory AppLogined({
    required String coreUrl,
    required String tenantId,
    required String token,
  }) = _AppLogined;
}

ycherniavskyi (approved):

PR #208, opened 2024-08-06
feat: implement call-to-action flow with action filtering by screen tab

SERDUN:


ycherniavskyi:
Task linked: [CU-86951n2ng Implement the "call to action" buttons as a standard feature of the app](https://app.clickup.com/t/86951n2ng)

PR #212, opened 2024-08-19
fix: correctly handle hangup events for Android background isolates

SERDUN:


ycherniavskyi:
Task linked: [CU-8694xtdy8 Cannot decline the call from pop-up notification on Android](https://app.clickup.com/t/8694xtdy8)

lib/background_call_handler.dart
```diff
@@ -53,27 +52,19 @@ class BackgroundCallHandler implements CallkeepBackgroundServiceDelegate {
...
-  Future _initializeSignalClient() async {
+  Future<void> _initializeSignalClient() async {
     final signalingUrl = _parseCoreUrlToSignalingUrl(storage.readCoreUrl() ?? '');
-
     final token = storage.readToken();
     final tenantId = storage.readTenantId() ?? '';
 
+    final certificates = AppCertificates().trustedCertificates;
     final httpClient = HttpClient();
     httpClient.connectionTimeout = kSignalingClientConnectionTimeout;
```
ycherniavskyi, line 62: It seems that this code is odd. It create local `httpClient` which is never used.
As I could see instead of this code lines we need to pass `connectionTimeout` property with  `kSignalingClientConnectionTimeout` value to `WebtritSignalingClient.connect` method call.

lib/background_call_handler.dart
```diff
@@ -105,65 +96,76 @@ class BackgroundCallHandler implements CallkeepBackgroundServiceDelegate {
...
-      );
-    } else if (event is IceHangupEvent) {
-      _callNotificationDelegate.endAllBackgroundCalls();
-    } else if (event is HangupEvent) {
-      _callNotificationDelegate.endBackgroundCall(event.callId);
-    } else if (event is DecliningEvent) {
-      _callNotificationDelegate.endBackgroundCall(event.callId);
-    } else if (event is UnregisteredEvent) {
-      _callNotificationDelegate.endAllBackgroundCalls();
+    _logger.warning('_onSignalingEvent $event');
```
ycherniavskyi, line 99: I am not sure this is `warning` message, more likely it is `info` message.

lib/background_call_handler.dart
```diff
@@ -105,65 +96,76 @@ class BackgroundCallHandler implements CallkeepBackgroundServiceDelegate {
...
+
+    final eventHandlers = <Type, void Function(Event)>{
+      IncomingCallEvent: (e) => _handleIncomingCall(e as IncomingCallEvent),
+      HangupEvent: (e) => _handleHangupCall(e as HangupEvent),
+      UnregisteredEvent: (e) => _handleUnregisteredEvent(e as UnregisteredEvent),
+    };
+
+    final handler = eventHandlers[event.runtimeType];
+    if (handler != null) {
```
ycherniavskyi, line 108: In my opinion, this type of logic is overengineered. Using `runtimeType` instead of `is` is less intuitive. Let's stick with the previous pattern using `if` and `is`. If you really want to write more advanced code, consider using [Dart Patterns](https://dart.dev/language/patterns), specifically [Switch statements and expressions](https://dart.dev/language/patterns#switch-statements-and-expressions).

lib/background_call_handler.dart
```diff
@@ -105,65 +96,76 @@ class BackgroundCallHandler implements CallkeepBackgroundServiceDelegate {
...
-  void _declineCall(String callId) async {
+  Uri _parseCoreUrlToSignalingUrl(String coreUrl) {
+    final uri = Uri.parse(coreUrl);
+    return uri.replace(scheme: uri.scheme.endsWith('s') ? 'wss' : 'ws');
+  }
+
+  Future<void> _declineCall(String callId) async {
     final transaction = WebtritSignalingClient.generateTransactionId();
-    var line = retrieveIdleLine(_lines, callId) ?? _kUndefinedLine;
+    final lineIndex = _lines.indexWhere((line) => line?.callId == callId);
```
ycherniavskyi, line 147: `line` already mean index, so `lineIndex` is overcomplicated naming which will conflict with other `line` variables names.

lib/background_call_handler.dart
```diff
@@ -196,11 +198,9 @@ class BackgroundCallHandler implements CallkeepBackgroundServiceDelegate {
...
   }) {
-    if (value.trim().toLowerCase() == 'true') {
-      return true;
-    } else if (value.trim().toLowerCase() == 'false') {
-      return false;
-    }
+    final trimmedValue = value.trim().toLowerCase();
+    if (trimmedValue == 'true') return true;
+    if (trimmedValue == 'false') return false;
     return defaultValue;
```
ycherniavskyi, line 204: Let's use `else if` structure with brackets here. It is easy to read and more friendly for adding code to some of conditions if needed in future.

ycherniavskyi (changes requested):

lib/background_call_handler.dart
SERDUN: Yes, you’re right. It’s odd, and I don’t even remember when I added it. Thanks!

lib/background_call_handler.dart
SERDUN: It appears that this function is not being used. I will remove it. Interesting, why didn't the linter highlight it?)

PR #213, opened 2024-08-20
feat: handle signaling response codes during hangup process

SERDUN:


ycherniavskyi:
Task linked: [CU-8695e03gg NovaSys: check if there is an issue: improve logout flow;  app is hanging on call after user hangs up](https://app.clickup.com/t/8695e03gg)

PR #214, opened 2024-08-21
feat: handle edge cases of incorrect call flow execution

SERDUN:
…is present


SERDUN:
Task linked: https://app.clickup.com/t/8695e03gg

lib/features/call/view/call_active_scaffold.dart
```diff
@@ -153,6 +153,10 @@ class CallActiveScaffoldState extends State<CallActiveScaffold> {
                             username: activeCall.displayName ?? activeCall.handle.value,
                             acceptedTime: activeCall.acceptedTime,
                             color: onTabGradient,
+                            callProcessingStatus: CallProcessingStatusView(
+                              status: activeCall.status,
+                              color: onTabGradient,
+                            ),
```
ycherniavskyi, line 159: Hmm, I am not agree with such solution for `CallInfo` widget. I mean make some tiny part of it as other widget container.
Let's get rig of `CallProcessingStatusView` widget completely and inject it into `CallInfo` widget and just pass `status` property in it.

lib/features/call/view/call_shell.dart
```diff
@@ -115,10 +113,37 @@ class _CallShellState extends State<CallShell> {
...
+    }
+  }
+
+  @override
+  void dispose() {
+    _removeThumbnailAvatar();
+    super.dispose();
+  }
+
+  void _openCallScreen(StackRouter router, CallState state) {
```
ycherniavskyi, line 131: Pass while call state as parameter is to wide here. Let's just use already used `final hasActiveCalls = state.activeCalls.isNotEmpty` local variable pattern and pass `bool hasActiveCalls` parameter instead of `state`.

lib/features/call/bloc/call_state.dart
```diff
@@ -127,6 +127,7 @@ class ActiveCall with _$ActiveCall {
     required String callId,
     required CallkeepHandle handle,
     String? displayName,
+    CallProcessingStatus? status,
```
ycherniavskyi, line 130: Let's rename `CallProcessingStatus` to `ActiveCallStatus`.

ycherniavskyi (changes requested):

lib/features/call/view/call_active_scaffold.dart
```diff
@@ -153,6 +154,16 @@ class CallActiveScaffoldState extends State<CallActiveScaffold> {
...
+                            callProcessingStatus: activeCall.status != null
+                                ? Padding(
+                                    padding: const EdgeInsets.only(top: 8),
+                                    child: Text(
+                                      activeCall.status!.l10n(context),
+                                      style: textTheme.labelLarge!.copyWith(color: onTabGradient),
+                                      textAlign: TextAlign.center,
+                                    ),
+                                  )
+                                : const SizedBox(),
```
ycherniavskyi, line 166: That is not what I mean. I propose to put all this widget related code in `CallInfo` widget and instead `callProcessingStatus` property pass `activeCallStatus` property here.

ycherniavskyi (approved):

PR #216, opened 2024-08-22
feat: add logic for database migration handling downgrade

SERDUN:
Perform a clean downgrade of the database when the app is downgraded from a newer test flight version to an older production version.
This ensures that the database schema is compatible with the older version by removing all existing tables.


SERDUN:
Task linked: https://app.clickup.com/t/8695g1yw6

PR #225, opened 2024-09-03
fix: prevent route duplication by switching to `navigate` for CallScreenPageRoute

SERDUN:


ycherniavskyi:
Task linked: [CU-86956pyy9 Fix navigation to active call](https://app.clickup.com/t/86956pyy9)

PR #229, opened 2024-09-12
feat: handle signaling code for SIP unavailable service

SERDUN:


ycherniavskyi:
Task linked: [CU-8695muz1x Strange bugs when logging in demo user](https://app.clickup.com/t/8695muz1x)

packages/webtrit_signaling/lib/src/signaling_registration_failed_code.dart
```diff
@@ -0,0 +1,16 @@
...
+
+  final int code;
+}
+
+extension SignalingRegistrationFailedCodeByCode on Iterable<SignalingRegistrationFailedCode> {
+  SignalingRegistrationFailedCode byCode(int code) {
+    for (var value in this) {
+      if (value.code == code) return value;
+    }
+    return SignalingRegistrationFailedCode.sipServerUnavailable;
```
ycherniavskyi, line 14: Here must be used some special code - `unmappedCode`  (with `code` equals `-1`). (see how it is implemented in `SignalingDisconnectCodeByCode`).

lib/features/call/bloc/call_bloc.dart
```diff
@@ -381,27 +381,41 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
...
-        notificationsBloc.add(NotificationsSubmitted(ErrorMessageNotification(event.reason!)));
-      } else {
-        notificationsBloc.add(NotificationsSubmitted(AppOfflineNotification()));
+    if (newRegistrationStatus.isUnregistered) {
+      notificationsBloc.add(NotificationsSubmitted(AppOfflineNotification()));
+      return;
+    }
+
+    if (newRegistrationStatus.isFailed) {
+      final code = SignalingRegistrationFailedCode.values.byCode(event.registration.code!);
```
ycherniavskyi, line 410: Use `!` is potentially problematic construction. Let's introduce `SignalingRegistrationFailedCode.nullCode` (with `code` equals `0`) and check to `null` before `byCode` call.

ycherniavskyi (changes requested):

lib/features/call/bloc/call_bloc.dart
```diff
@@ -996,35 +1006,35 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
...
-    add(const _RegistrationChange(registrationStatus: RegistrationStatus.registered));
+    add(const _RegistrationChange(registration: Registration(status: RegistrationStatus.registered)));
   }
 
   Future<void> __onCallSignalingEventRegistrationFailed(
     _CallSignalingEventRegisterationFailed event,
     Emitter<CallState> emit,
   ) async {
-    add(const _RegistrationChange(registrationStatus: RegistrationStatus.registration_failed));
+    add(const _RegistrationChange(registration: Registration(status: RegistrationStatus.registration_failed)));
```
ycherniavskyi, line 1023: It seems that we need to extend `_CallSignalingEvent.registrationFailed()` (which is corespondent to `_CallSignalingEventRegisterationFailed`) constructor with `code` and `reason` parameters and populate it during  `RegistrationFailedEvent` mapping in method `_onSignalingEvent`.
It was missed previously, but I propose to do it within this PR.

ycherniavskyi (approved):

PR #235, opened 2024-09-18
feat: add settings embedded pages

SERDUN:


lib/features/settings/view/settings_screen.dart
```diff
@@ -83,121 +82,28 @@ class SettingsScreen extends StatelessWidget {
...
-            leading: const Icon(Icons.delete_outline),
-            title: Text(context.l10n.settings_ListViewTileTitle_accountDelete),
-            onTap: () async {
-              final settingsBloc = context.read<SettingsBloc>();
-              final deleteAccount = await ConfirmDialog.show(
-                context,
-                title: context.l10n.settings_AccountDeleteConfirmDialog_title,
-                content: context.l10n.settings_AccountDeleteConfirmDialog_content,
+          Column(
+            children: sections.map((section) {
```
ycherniavskyi, line 86: Let's use [*collection for*](https://dart.dev/language/collections#control-flow-operators) here instead of map. ()

lib/features/settings/view/settings_screen.dart
```diff
@@ -212,4 +118,37 @@ class SettingsScreen extends StatelessWidget {
       child: scaffold,
     );
   }
+
+  void _handleNavigation(BuildContext context, SettingItem item) {
```
ycherniavskyi, line 122: It seems that within this method not only navigation executed (I mean `_deleteAccount` method call), so I propouse to rename it to more general name - `_onSectionItemTap` for example.

lib/models/feature_access/config_data.dart
```diff
@@ -0,0 +1,22 @@
...
+      url: Uri.parse(url),
+    );
+  }
+
+  factory ConfigData.create(String url, String titleL10n) {
+    return ConfigData(
+      url: Uri.parse(url),
+      titleL10n: titleL10n,
+    );
+  }
```
ycherniavskyi, line 21: What is the reason for this factories? As for me one contractor will be enough.

lib/models/feature_access/config_data.dart
```diff
@@ -0,0 +1,22 @@
+class ConfigData {
+  final Uri url;
+  final String? titleL10n;
```
ycherniavskyi, line 3: As I remember we declare fields after contractors.

ycherniavskyi (changes requested):

PR #240, opened 2024-09-24
feat: add option to enable/disable adding contacts to favorites

SERDUN:


lib/features/contact/widgets/contact_phone_tile.dart
```diff
@@ -11,10 +11,7 @@ class ContactPhoneTile extends StatelessWidget {
     required this.transfer,
     this.onTap,
     this.onLongPress,
-    this.onFavoriteChanged,
-    this.onAudioPressed,
-    this.onVideoPressed,
-    this.onTransferPressed,
+    required this.trailingIcons,
```
ycherniavskyi, line 14: Hmm, I don't like such modification, because we extract internal logic of `contact_phone_tile` to external widgets and convert it in just container. That is why I propose to just add `isFavoritesAvailable` and `isTransferAvailable` with `true` default value.

lib/features/contact/widgets/contact_phone_tile.dart
```diff
@@ -2,11 +2,14 @@ import 'package:flutter/material.dart';
...
+// Follow naming conventions as outlined in https://api.flutter.dev/flutter/widgets/Visibility-class.html
 class ContactPhoneTile extends StatelessWidget {
   const ContactPhoneTile({
     super.key,
     required this.number,
     required this.label,
+    required this.favoritesVisible,
+    required this.transferVisible,
     required this.favorite,
     required this.transfer,
```
ycherniavskyi, line 14: it seems that we need to use `favoriteVisible` naming for added property to be consist with `favorite` property. And actually `favoriteVisible` is more reasonable for `ContactPhoneTile` and `ContactScreen` widgets.

PR #245, opened 2024-09-26
feat: add session cleanup worker

SERDUN:


lib/app/router/main_shell.dart
```diff
@@ -109,6 +109,7 @@ class _MainShellState extends State<MainShell> {
         RepositoryProvider<UserRepository>(
           create: (context) => UserRepository(
             webtritApiClient: context.read<WebtritApiClient>(),
+            queueRequestWorker: SessionCleanupWorker(),
```
ycherniavskyi, line 112: Let's sync parameter naming with class name.

ycherniavskyi, line 112: And also let's move this parameter before `webtritApiClient`.

lib/repositories/account/user_repository.dart
```diff
@@ -12,9 +14,11 @@ final _logger = Logger('UserRepository');
 class UserRepository {
   UserRepository({
     required WebtritApiClient webtritApiClient,
+    required SessionCleanupWorker queueRequestWorker,
```
ycherniavskyi, line 17: Let's move this parameter before `webtritApiClient`. And let's make it optional when in case of `null` value this functionality is disabled.

packages/webtrit_api/lib/src/exceptions.dart
```diff
@@ -2,13 +2,15 @@ import 'models/error.dart';
...
-    required this.statusCode,
+    required this.tenantUrl,
+    this.statusCode,
     required this.requestId,
     this.token,
     this.error,
   });
 
-  final int statusCode;
+  final Uri tenantUrl;
```
ycherniavskyi, line 12: As for me for general `RequestFailure` class this property will be better to name `url`.

lib/data/session_cleanup_worker.dart
```diff
@@ -0,0 +1,140 @@
...
+  static const String uriKey = 'uri';
+  static const String tokenKey = 'token';
+  static const String timestampKey = 'timestamp';
+  static const String attemptsKey = 'attempts';
+  static const String maxAttemptsKey = 'maxAttempts';
+  static const String expirationKey = 'expiration';
+
+  Future<SharedPreferences> get _prefs async => await SharedPreferences.getInstance();
+
+  Future<void> storeFailedSession(Uri uri,
```
ycherniavskyi, line 95: Let's use `url` naming instead `uri` within this class (for this variable and for key variable name and its value).

ycherniavskyi (changes requested):

ycherniavskyi (approved):

PR #251, opened 2024-10-01
feat: improve login flow

SERDUN:


lib/features/login/features/login_embedded/view/login_embedded_screen_page.dart
ycherniavskyi, line 1: It is a bit odd to have `LoginEmbeddedScreenPage` (with `LoginEmbedded` prefix) and `EmbeddedLogin` classes in one place related to one functionality.
Could we choose one order from `LoginEmbedded` and `EmbeddedLogin` combinations?

ycherniavskyi (changes requested):

lib/theme/models/app_config.dart
```diff
@@ -54,8 +54,9 @@ class AppConfigLoginEmbedded with _$AppConfigLoginEmbedded {
 
   const factory AppConfigLoginEmbedded({
     required int id,
-    required bool launch,
-    required String titleL10n,
+    @Default(false) bool launch,
+    @Default('') String titleL10n,
```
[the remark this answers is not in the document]
SERDUN: I was thinking about the showToolbar flag. If showToolbar=false, then titleL10n is not necessary, and we can omit it from the schema and skip the additional nullable check in the code

lib/features/login/features/login_embedded/view/login_embedded_screen_page.dart
SERDUN: Lets use EmbeddedLogin for order starts with the more general concept (Embedded) followed by the specific feature (Login), which can make it easier to understand that it’s an embedded version of login functionality.

ycherniavskyi (approved):

PR #254, opened 2024-10-23
feature: implement foreground service for reliable VoIP call handling

SERDUN:
Add option to select notification type for incoming calls or background socket handling


SERDUN:
https://app.clickup.com/t/86964vjz4

PR #257, opened 2024-10-24
feat: add the possibility to use core URL from auto-provisioning

SERDUN:


SERDUN:
https://app.clickup.com/t/869614wp4

PR #260, opened 2024-10-29
fix: handle nullable `notificationsService` in _MessagingShellState to prevent disposal errors

SERDUN:


ycherniavskyi:
Task linked: [CU-8696dp3pt Widget MainShell is not dispose in develop branch](https://app.clickup.com/t/8696dp3pt)

PR #325, opened 2024-11-28
fix: ensure backward compatibility with previous core version in contacts flow

SERDUN:


ycherniavskyi:
Task linked: [CU-8696vjj6x Provide backward compatibility with previous COREs in contacts](https://app.clickup.com/t/8696vjj6x)

PR #328, opened 2024-12-03
refactor: centralize log initialization into a single data class for cross-isolate usage

SERDUN:
•	Consolidated log initialization into a single data class to avoid code duplication across different places.
•	Enhanced flexibility by enabling logic extensions for filtering and anonymizing logs.
•	Improved handling of multiple data sources for better scalability and maintainability.


digiboridev (approved):
yes

PR #346, opened 2024-12-16
feat:  synchronize local connections with core

SERDUN:
Pleas check first: https://github.com/WebTrit/webtrit_callkeep/pull/49

This update handles cases where user actions on connections (e.g., accept or decline) don’t reach the server, potentially due to network issues. It provides proper handling for incoming, outgoing, and accepted calls, ensuring the application state remains consistent even if the server does not acknowledge these actions immediately.


ycherniavskyi:
Task linked: [CU-8696vhn6g "Call request on busy line" error](https://app.clickup.com/t/8696vhn6g)

PR #350, opened 2024-12-16
feat: enable use of app session identifier in remote logging system

SERDUN:
- Move app ID retrieval to a dedicated App ID provider, allowing the application to function without Firebase when not needed and accommodating cases where Google services are unavailable.

- Integrate the app session identifier to enable targeted filtering in the remote logging system.


digiboridev (approved):
gj

PR #368, opened 2024-12-30
feat: migrate from withOpacity to withValues(alpha)

SERDUN:
https://docs.flutter.dev/release/breaking-changes/wide-gamut-framework#opacity


digiboridev (approved):
gj

PR #378, opened 2025-01-14
refactor: in app notifications beautify

digiboridev:
Contains various changes to ensure predictable and understandable app error handling in same manner for all app conponents to deliver it to user throught ui.


SERDUN (approved):
Ok

PR #381, opened 2025-01-15
refactor(messaging): edits after preproduction tests 2025.01.14

digiboridev:
Contains little bunch of various improovements and fixes, according to last manual test for preproduction.


SERDUN (approved):
Ok

PR #383, opened 2025-01-16
refactor: contact db structure

digiboridev:
Contains refactoring of legacy contacts data structure, that leads to use contact entity separately of their phones and emails so its produce bugs, inconsistency and confusing.


SERDUN (approved):
Ok

PR #385, opened 2025-01-16
fix: contact screen avatar layout

digiboridev:
Contains fix for broken avatar layout on contact screen due to unconstrained width and adds better support for scaling.


SERDUN (approved):
Ok

PR #387, opened 2025-01-16
feat: add contacts agreement

SERDUN:
Added contacts agreement functionality for collecting contacts as requested by Google reviewers. If the user does not agree to the collection of contacts, permissions are not requested, and no data is stored in the database.

Points:

- Added a model to represent the status of the agreement. Consider applying the same approach for user agreements instead of using a bool for consistency in the code.


digiboridev (approved):
ok

PR #390, opened 2025-01-19
feat: allow changing contacts agreement choice on the diagnostic screen

SERDUN:
- Refactored the agreement checkbox widget for better reusability and clarity.
- Enabled the ability to change the agreement choice directly from the diagnostic screen.


digiboridev (approved):
ok

PR #391, opened 2025-01-19
feat: include tenant ID from embedded signup flow in signup request

SERDUN:
- Include tenant ID from the embedded signup extras response in the signup request.
- Add documentation for configuring the signup login process.


digiboridev (approved):
ok

PR #392, opened 2025-01-20
fix: avatar layered to hierarhical layout

digiboridev:
Change avatar lyout from layered to hierarhical to avoid bug with transparent images and improove memory consumption


SERDUN (approved):
Ok

PR #395, opened 2025-01-22
feat: add conditional WebView support for web platform using `kIsWeb`

SERDUN:
Enable support for screens with WebView in the configurator


digiboridev (approved):
ok

PR #397, opened 2025-01-28
fix: multiline call drop if active call held'nt

digiboridev:
Fix for multi line calls thats drops active call if it not hold before making new call


SERDUN (approved):
Ok

PR #401, opened 2025-02-03
feat: encoding settings

digiboridev:
Contains implementation for *"media encoding settings"* feature. 

That provides ability to fine tune and improove media encoding via sdp modification.
Abilities:
- set audio/video bitrate
- set packetization time (rtp chunk size)
- set frequency bandwidth
- set stereo mode
- set dts mode (silence discart)
- reordering(prioritizing) and exclusion rtp profiles


SERDUN (approved):
Ok

PR #403, opened 2025-02-05
refactor: data class dependency injection

SERDUN:
- Use inheritance to provide the data class instead of using singleton access.
- Use constructor injection to enable mocking of data.
- Refactor PackageInfo to allow data mocking.


digiboridev (approved):
gj

PR #405, opened 2025-02-05
refactor: update device info and provide user agent via utility

SERDUN:
- Added an interface for DeviceInfo to enable mock data creation for testing
- Provided DeviceInfo via an InheritedWidget instead of accessing it through a singleton
- Moved user agent handling to the WebView constructor and a utility function for generating it based on the DeviceInfo data class


digiboridev (approved):
ok

PR #418, opened 2025-02-13
feat: call processing refactoring and status ui

digiboridev:
- Extend and refactoring call processing statuses
- Improoving call race conditions handling using new statuses
- Add ui representation for new statuses


SERDUN (approved):
Ok

PR #420, opened 2025-02-17
feat: add global scope for embedded resources to simplify access

SERDUN:
- Store embedded pages in a global scope instead of nested objects, adding ID relationships to avoid duplication and simplify access.
- Correctly name the resource URI property for better clarity.
- Fix missing embedded title in the settings section.


digiboridev (approved):
ok

PR #421, opened 2025-02-17
feat: encoding presets and env overrides

digiboridev:
Adds support for predefined media encoding presets with optimized default for voip (according to rfc recomended sweet spots, and following other top messengers configs investigation).
Supports external override for easy configurator integration if will needed.


SERDUN (approved):
Ok

PR #422, opened 2025-02-17
refactor: improve URI handling and simplify embedded resources management

SERDUN:
- extract URI parsing logic into a separate class for reuse with other asset resources


digiboridev (approved):
ok

PR #437, opened 2025-02-24
feat: improuve log flow

SERDUN:
- prevent emitting crashes in the logzio http client when the system restricts internet access
- add logs for the fcm isolate
- prevent emitting debug logs in crashlytics


digiboridev (approved):
ok

PR #455, opened 2025-03-03
feat: add logging for failed registration status with error code and reason

SERDUN:
Add missing log for failed registration status with error code and reason


digiboridev (approved):
ok

PR #487, opened 2025-04-03
fix: fallback to video flag when remote stream is unavailable to indicate original incoming call type

SERDUN:
Allows showing video icon for incoming calls even before the remote stream is available

Note: switching from video to audio before answer is not yet handled and requires signaling update


ycherniavskyi:
Task linked: [CU-8698jmky4 There is an audio button while accepting a video call](https://app.clickup.com/t/8698jmky4)

PR #488, opened 2025-04-09
fix: add inactive local video track before createAnswer for compatibility with certain core instances

SERDUN:
Adds local inactive video track before createAnswer to activate the transceiver. Enables configuration option to control this behavior, with a default value provided. Improves compatibility with servers where callee track stays inactive without local track.


digiboridev (approved):
Gj

PR #499, opened 2025-04-22
feat: add support for injecting payload data into WebView

SERDUN:
Extend the schema with payload data to define which information should be injected into the WebView.
Some resources may come from local storage, while others (e.g., external page token) need to be prepared in advance.


digiboridev (approved):
ok

PR #551, opened 2025-05-30
fix: navigation bar overlap on auth screens

JohnBorys:


ycherniavskyi:
Task linked: [CU-86998e5p2 Fix Navigation Bar Overlap on Auth Screens](https://app.clickup.com/t/86998e5p2)

PR #552, opened 2025-05-30
feat: add version-based flavor selection system with backward compati…

SERDUN:
…bility


ycherniavskyi:
Task linked: [CU-86998e95w add versioned config and extract shared build logic to Makefile.shared](https://app.clickup.com/t/86998e95w)

PR #553, opened 2025-05-31
feat: add support for SMS trigger to initiate incoming call

SERDUN:
Added a new smsReceiver flavor for Android to support incoming call initialization via SMS. The UI has been updated to reflect the current status of the receiver (enabled/disabled), and runtime permission handling has been implemented.

Flavors are currently added only for Android, as iOS support is not critical at this stage. However, this introduces some complexity when running and building different targets.

Makefiles have been updated to automatically apply flavors based on the VERSION defined in build.config. This ensures backward compatibility for older versions that do not support flavors. When building APKs or app bundles, the appropriate flavors are included automatically. For iOS, flavor usage is explicitly disabled.

There might still be some inconvenience when debugging during development, as flavors need to be manually specified in IDE configurations or passed via a Makefile target. I’ll create a separate pull request with documentation on how to build and run using flavors. https://github.com/WebTrit/webtrit_phone/pull/555

### DIFF
--- a/lib/features/settings/features/network/view/network_screen.dart
+++ b/lib/features/settings/features/network/view/network_screen.dart
@@ -31,9 +31,8 @@ class NetworkScreen extends StatelessWidget {
               mainAxisAlignment: MainAxisAlignment.start,
               mainAxisSize: MainAxisSize.min,
               children: [
-                Padding(
-                  padding: const EdgeInsets.all(16.0),
-                  child: Text(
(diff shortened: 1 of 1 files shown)


ycherniavskyi:
Task linked: [CU-86998j5xw Add possibility to disable/enable sms trigger of incomming call](https://app.clickup.com/t/86998j5xw)

PR #566, opened 2025-06-16
fix: system notificaitons badge colour

digiboridev:


digiboridev:
> ## Pull Request Overview
> This pull request fixes the system notifications badge colour by setting the Material widget's color to transparent so that it no longer overrides the intended styling.
> 
> * Fixes badge colour rendering by setting Material color to transparent.

alt text

PR #583, opened 2025-06-26
feat: add support for opening external links from WebView

SERDUN:
Introduces a custom URI scheme (app://openInExternalBrowser?url=...) to explicitly mark links that should be opened outside of the WebView using the system browser. This improves link handling by separating internal navigation from external actions.

### DIFF
--- a/docs/embedded_pages.md
+++ b/docs/embedded_pages.md
@@ -2,7 +2,9 @@
 
 ## Overview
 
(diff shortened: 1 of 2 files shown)


ycherniavskyi:
Task linked: [CU-8699j7514 Support iOS Self-Service Flow via Browser Redirect from Web App](https://app.clickup.com/t/8699j7514)

PR #586, opened 2025-07-02
feat: show reminder for persistent connection on Android 14+

ivan-bilchenko:


lib/features/settings/features/network/view/network_screen.dart
```diff
@@ -80,4 +97,32 @@ class NetworkScreen extends StatelessWidget {
       ),
     );
   }
+
+  Future<bool> _isAndroidVersionAtLeast(int targetVersion) async {
```
SERDUN, line 101: Maybe it’s better to extend DeviceInfo and extract common logic there, such as Android version or range.

PR #605, opened 2025-07-23
fix: the error occurred due to a race condition in the onRenegotiation

ivan-bilchenko:
Root Cause:
The error occurred due to a race condition in the onRenegotiationNeeded callback, where a local offer was attempted while the peer connection was in the have-remote-offer state, which is invalid according to WebRTC rules.

Solution:
Implemented a state-aware renegotiation queue system. All renegotiation requests are now queued and processed only when the signaling state is stable. This prevents illegal operations and ensures safe, reliable renegotiation.


ycherniavskyi:
Task linked: [CU-86997yj9b Video call transfer to number without video fails](https://app.clickup.com/t/86997yj9b)

PR #713, opened 2025-10-31
fix: show contact extension only when available

SERDUN:
This PR adds localized support for displaying contact extension numbers in the messaging conversation builder. The hardcoded "Ext:" label has been replaced with localizable strings, and the logic has been improved to hide the subtitle when no extension is available.

Added localization entries for contact extension labels in English, Italian, and Ukrainian
Updated generated localization files to include the new translation method
Refactored the contact list UI to use localized strings and conditionally display the extension


ycherniavskyi:
Task linked: [CU-869az3f87 Uknown extensions in chat list](https://app.clickup.com/t/869az3f87)

PR #726, opened 2025-11-05
refactor: make registration nullable, unify events, and allow early outgoing calls

SERDUN:
This PR changes the registration field in CallServiceState from non-nullable to nullable (Registration?). This allows the registration status to be null before the signaling handshake is fully initialized, particularly when a call is initiated before the app is fully active (e.g., from phone recents).

Key changes:

- Made the registration field nullable in CallServiceState and its generated freezed code

- Consolidated multiple registration-related signaling events into a single _CallSignalingEventRegistration event

- Added null-safe checks throughout the codebase where registration is accessed

- Introduced isHandshakeEstablished and isSignalingEstablished getters to track connection state

- Enhanced continueStartCallIntent to wait for both signaling and handshake establishment before proceeding with a call


SERDUN:
Migrates part of the fix from [#724](https://github.com/WebTrit/webtrit_phone/pull/724)￼

PR #732, opened 2025-11-10
fix: show notification when self-managed phone account registration fails

SERDUN:
This PR adds localized error handling for a self-managed phone account registration failure scenario. When the app fails to register a self-managed phone account with the system, users will now see an appropriate error message in their preferred language.

Added new localization strings for Ukrainian, Italian, and English languages
Created a new notification class to display the error message
Updated call handling logic to show the notification when registration fails

Related: https://github.com/WebTrit/webtrit_callkeep/pull/105


ycherniavskyi:
Task linked: [CU-869b1vfxz Getting 'Preparing' while trying to call](https://app.clickup.com/t/869b1vfxz)

digiboridev (approved):

PR #746, opened 2025-11-19
feat: disable FCM token retry

OMaudza:
Cancel FCM token retry when Google Play Services is missed or disabled


SERDUN:
Actual PR: https://github.com/WebTrit/webtrit_phone/pull/771/files

PR #764, opened 2025-11-26
fix: handle missing microphone permission in call

OMaudza:
Handle missing microphone permission in incoming call. Also add warning indicator on Settings screen


ycherniavskyi:
Task linked: [CU-869a8krng Handle missing microphone permission during incoming call + add in-app warning indicator](https://app.clickup.com/t/869a8krng)

digiboridev (approved):

PR #817, opened 2025-12-24
fix: not available media

OMaudza:
Fixed bug with no available media for 20 seconds in Firefox after enabling video.


ycherniavskyi:
Task linked: [CU-869bgcgu7 No media for 20sec after enabling video on app side](https://app.clickup.com/t/869bgcgu7)

PR #836, opened 2026-01-02
fix fix changing info icon color

OMaudza:
Fixed bug when info icon color changes when user opens and closes menu section in media settings.

### DIFF
--- a/lib/features/settings/features/media_settings/widgets/heading_section.dart
+++ b/lib/features/settings/features/media_settings/widgets/heading_section.dart
@@ -9,6 +9,7 @@ class HeadingSection extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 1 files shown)


lib/features/settings/features/media_settings/widgets/heading_section.dart
```diff
@@ -24,7 +24,7 @@ class HeadingSection extends StatelessWidget {
             padding: const EdgeInsets.all(16),
             margin: const EdgeInsets.all(16),
             showDuration: const Duration(seconds: 10),
-            child: const Icon(Icons.info_outline),
+            child: const Icon(Icons.info_outline, color: Colors.black54,),
```
digiboridev: Use colors from color scheme please, and try re-format code because comma after color looks imposible for single line definition

PR #866, opened 2026-01-19
fix: add additional check for setting source object

OMaudza:
Added additional check when setting source object for renderer to prevent error when call initialize before setting the stream

### DIFF
--- a/lib/features/call/widgets/stream_thumbnail.dart
+++ b/lib/features/call/widgets/stream_thumbnail.dart
@@ -34,6 +34,8 @@ class _StreamThumbnailState extends State<StreamThumbnail> {
   final RTCVideoRenderer _renderer = RTCVideoRenderer();
   bool _isRendererInitialized = false;
 
+  bool get _canRender => _isRendererInitialized && _renderer.textureId != null;
+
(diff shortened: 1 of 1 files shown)


ycherniavskyi:
Task linked: [CU-869a52h2x _CallActiveThumbnailState.dispose](https://app.clickup.com/t/869a52h2x)

lib/features/call/widgets/stream_thumbnail.dart
```diff
@@ -55,7 +55,7 @@ class _StreamThumbnailState extends State<StreamThumbnail> {
   }
 
   void _updateSrcObject() {
-    if (_isRendererInitialized) {
+    if (_isRendererInitialized && _renderer.textureId != null) {
```
SERDUN: Let's move the logic into a private getter, for example: bool get _canRender => _isRendererInitialized && _renderer.textureId != null;

SERDUN:
Also, please use _canRender in dispose(), since RTCVideoRenderer can throw an exception if the texture ID is null.

SERDUN (approved):

PR #867, opened 2026-01-19
fix: fixed delaying when updating seen status

OMaudza:
Fixed delaying when changing seen status flag for voicemails


ycherniavskyi:
Task linked: [CU-869bn3nkx [WT] WT-548: The voicemail 'Seen' flag in the app changes with a delay](https://app.clickup.com/t/869bn3nkx)

SERDUN:
Please add tests that cover the fixed case and the expected behavior: the test should fail without the fix and pass with the fix. Also add another test for a common scenario.

SERDUN (approved):

PR #869, opened 2026-01-19
fix: change ice in media settings to abbreviation

OMaudza:
In media settings user can see ice and Ice. It was changed to ICE because it is abbreviation. Also settings_iceSettings_Section_title and settings_iceSettings_Section_tooltip in Italian language was changed because with abbreviation meaning was changed.


ycherniavskyi:
Task linked: [CU-869bhb172 PortaPhone internal: term "ice" should be in capital "ICE"](https://app.clickup.com/t/869bhb172)

PR #870, opened 2026-01-20
fix: prevent keepalive write-after-close and harden transaction lifecycle

OMaudza:
## Problem

Race condition between WebSocket keepalive timer and socket close event in Dart's event loop. When the connection drops, both events are queued — if the timer fires before `onDone`, it writes to a dead socket, causing an unhandled `StateError` that propagates to `_RawReceivePort._handleMessage` (337 crashes, Feb 21–28).

```
 Event loop queue:
 ┌─────────────────┐   ┌─────────────────┐
 │ keepalive timer  │   │ socket onDone   │
 │ (fires first)    │──▶│ (arrives later) │
 └────────┬────────┘   └────────┬────────┘
          │                     │
    sink.add() on               stops timer,
    dead socket                 runs disconnect
          │
    StateError ──▶ _onError ──▶ endCallsOnService()
                                ↑ all active calls killed
```

### 1. Pre-check `closeCode` before write (`_addData`)
Skip the write if the socket is already known to be closed. Wrap `StateError` into typed `WebtritSignalingBadStateException`.

### 2. Graceful keepalive stop (`_onKeepalive`)
- Catch `WebtritSignalingBadStateException` → silently exit the loop (no `_onError`)
- Check `closeCode` after successful keepalive → don't restart timer if socket closed mid-transaction

### 3. Transaction cleanup on send failure (`_executeTransaction`)
Moved `_addMessage` inside try-catch so the transaction is removed from the map if the write fails. Previously it leaked — stayed in `_transactions` forever with no response coming.

## Why silent return is safe

Reconnection is triggered by `Connectivity().onConnectivityChanged`, not by `_onError` or `_onDisconnect`. Suppressing the error only prevents unnecessary call termination — it does not block reconnection.

## Tests

5 tests covering both scenarios:

**Keepalive race condition** (fakeAsync)
- Timer fires on closed socket → graceful stop, `onError` not called
- `closeCode` set after successful keepalive → timer not restarted

**Transaction cleanup on send failure** (async)
- `sink.add` throws `StateError` → `WebtritSignalingBadStateException`
- `closeCode` already set → `WebtritSignalingBadStateException`
- `closeCode` detected before write → `sink.add` never called

### DIFF
--- a/packages/webtrit_signaling/lib/src/exceptions.dart
+++ b/packages/webtrit_signaling/lib/src/exceptions.dart
@@ -52,6 +52,15 @@ class WebtritSignalingTransactionTimeoutException extends WebtritSignalingTransa
   const WebtritSignalingTransactionTimeoutException(super.id, super.transactionId);
 }
 
+class WebtritSignalingBadStateException extends WebtritSignalingException {
+  const WebtritSignalingBadStateException(super.id, this.error);
+
+  final StateError error;
+
+  @override
+  String toString() => '${super.toString()}, stateError: $error';
+}
+
 class WebtritSignalingKeepaliveTransactionTimeoutException extends WebtritSignalingTransactionTimeoutException {
   const WebtritSignalingKeepaliveTransactionTimeoutException(super.id, super.transactionId);
 }
--- a/packages/webtrit_signaling/lib/src/webtrit_signaling_client.dart
+++ b/packages/webtrit_signaling/lib/src/webtrit_signaling_client.dart
@@ -271,7 +271,15 @@ class WebtritSignalingClient {
   void _addData(dynamic data) {
     _logger.fine(() => '_addData add: $data');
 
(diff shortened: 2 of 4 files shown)


ycherniavskyi:
Task linked: [CU-869a52d81 _RawReceivePort._handleMessage](https://app.clickup.com/t/869a52d81)

SERDUN:
I've pushed changes that seem closer to covering the exception reported in Crashlytics.
However, I'm not fully confident in this fix yet, as I haven't been able to clearly reproduce the issue or fully understand the reproduction flow.
The next step should be to reliably reproduce the bug - once we have that, we can confirm whether these changes actually address the crash or if the logic needs to be adjusted.

SERDUN:
Please check the logs in Logz.io and review the reported cases. You can use filters to search for specific logs.

PR #872, opened 2026-01-22
Fix/cu 869bn3nkx wt wt 548 the voicemail seen flag in the app changes with a delay v1

SERDUN:


ycherniavskyi:
Task linked: [CU-869bn3nkx [WT] WT-548: The voicemail 'Seen' flag in the app changes with a delay](https://app.clickup.com/t/869bn3nkx)

PR #876, opened 2026-01-23
fix: reset iOS audio route on call session start/end to prevent sticky speaker

SERDUN:
This PR fixes the "sticky speaker" issue on iOS where the audio route persists from previous call sessions, causing new calls to unexpectedly play through the speaker instead of the earpiece even when headphones are connected.

**Changes:**
- Removed obsolete commented-out code from `user_media_builder.dart` that was forcibly overriding audio routes
- Added lifecycle hooks in `call_bloc.dart` to reset iOS audio route to the receiver (earpiece) at the start of the first call and end of the last call


ycherniavskyi:
Task linked: [CU-869bwgzzu Speakerphone Toggles Randomly During Call](https://app.clickup.com/t/869bwgzzu)

digiboridev (approved):

PR #895, opened 2026-01-29
fix: solve "sticky speaker" issue by localizing audio state

SERDUN:
Moved `speakerOnBeforeMinimize` from global `CallState` to `ActiveCall` model to ensure audio preferences are call-specific and properly disposed of when a call ends.

- Fixes a bug where a new audio call would erroneously start on speakerphone if a previous call (or transfer session) had speaker enabled.
- Updated `CallBloc` to preserve and restore speaker state using the specific `ActiveCall` instance during minimization, transfers, and screen transitions.
- Cleaned up `CallState` to prevent stale audio state leakage between unrelated call sessions.


ycherniavskyi:
Task linked: [CU-869by7b3g Loudspeaker is active for all calls after user transfers a video call](https://app.clickup.com/t/869by7b3g)

digiboridev (approved):

PR #898, opened 2026-01-29
fix: capture CallBloc reference to prevent unsafe ancestor lookup

SERDUN:
This PR fixes a `FlutterError: Looking up a deactivated widget's ancestor is unsafe` that occurred during call transfers when accessing `context.read<CallBloc>()` inside callback closures. The fix captures the `CallBloc` reference at the start of the `build` method and uses this captured reference throughout all event handlers.

**Changes:**
- Captured `CallBloc` reference at build method start to prevent unsafe ancestor lookup
- Replaced all `context.read<CallBloc>()` calls with the captured bloc reference in callback closures
- Improved code consistency and formatting in several callback handlers

### DIFF
--- a/lib/features/call/view/call_active_scaffold.dart
+++ b/lib/features/call/view/call_active_scaffold.dart
@@ -76,6 +76,8 @@ class CallActiveScaffoldState extends State<CallActiveScaffold> {
 
     final style = themeData.extension<CallScreenStyles>()?.primary;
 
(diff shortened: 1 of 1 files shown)


ycherniavskyi:
Task linked: [CU-869bythj8 Poupup menu ignoring tap in some cases](https://app.clickup.com/t/869bythj8)

PR #904, opened 2026-02-02
fix: fixed contacts ordering

OMaudza:
Added collate method with no case option when sorting contacts


ycherniavskyi:
Task linked: [CU-869bvbuth [WT] WT-565: Contacts diverge from alphabetic order in PortaPhone Internal](https://app.clickup.com/t/869bvbuth)

SERDUN (approved):

PR #935, opened 2026-02-17
fix: fixed return button not working

OMaudza:
Fixed bug when return button does not work on Terms & Conditions screen in settings. Problem occurs when can not load web page.


SERDUN:
Need to clarify for not affect another flow

PR #936, opened 2026-02-17
fix: fixed localization issues in network settings

OMaudza:
Fixed issues with localization in network settings. Also added radio buttons and checkbox for 'Incoming Call Type' and 'SMS Fallback' accordingly


ycherniavskyi:
Task linked: [CU-869bw3ekc [WT] WT-566: Ukrainian 'Network settings' localizations issues in P1 Internal](https://app.clickup.com/t/869bw3ekc)

PR #990, opened 2026-03-04
fix: camera access exception on android

OMaudza:
Fixed CameraAccessException on Android. This error happens when user has two active video calls. When app calls build method of UserMediaBuilder for second video call, this exception is thrown. To fix this error additional checks for video tracks were added.


SERDUN:
Closing this one in favour of the proper fix at the native layer —
[WebTrit/flutter-webrtc#6](https://github.com/WebTrit/flutter-webrtc/pull/6) addresses the root cause
(`Camera2` being opened twice concurrently) without the risks around shared `MediaStream` ownership
or fragile index-based call lookup.

PR #999, opened 2026-03-12
fix: guard _handleRenegotiationNeeded against non-stable signaling state

SERDUN:
## Summary

- Replace `pcState != null` guard with `pcState == RTCSignalingState.RTCSignalingStateStable` to prevent `setLocalDescription` from being called in `have-remote-offer` or other non-stable states
- Add TOCTOU re-check after the async `createOffer({})` gap — if state changed in the meantime, log and skip `setLocalDescription`
- Remove outdated TODO comment that described this exact bug

## Problem

`_handleRenegotiationNeeded` was proceeding whenever `signalingState != null`, which includes `have-remote-offer`. This caused:

```
WEBRTC_SET_LOCAL_DESCRIPTION_ERROR: Failed to set local offer sdp: Called in wrong state: have-remote-offer
```

Triggered notably when `CalleeVideoOfferPolicy.includeInactiveTrack` fires `onRenegotiationNeeded` before the remote offer is processed.

## Test plan

- [x] Hold/unhold during active call — no crash
- [x] Mute/unmute video with `includeInactiveTrack` policy — renegotiation completes correctly
- [ ] Logs confirm `onRenegotiationNeeded signalingState` skips when state is `have-remote-offer`
- [ ] No regression on outgoing/incoming calls

### DIFF
--- a/lib/features/call/bloc/call_bloc.dart
+++ b/lib/features/call/bloc/call_bloc.dart
@@ -95,6 +95,7 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
   Timer? _presenceInfoSyncTimer;
 
   late final PeerConnectionManager _peerConnectionManager;
+  late final RenegotiationHandler _renegotiationHandler;
 
   final _callkeepSound = WebtritCallkeepSound();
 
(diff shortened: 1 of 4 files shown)


SERDUN:
Closing in favour of a consolidated PR that includes the ICE restart guard, renegotiation guard, and glare condition fix together.

PR #1054, opened 2026-04-02
fix(WT-1083): use processingStatus to guard outgoing call reconciliation

SERDUN:
## Summary

- Fixes a bug where outgoing calls with a failed `OutgoingCallRequest` would get permanently stuck during `StateHandshake` reconciliation
- Replaces the broad `acceptedTime == null && hungUpTime == null` skip with a precise check against `processingStatus`
- No new fields added — `CallProcessingStatus.outgoingOfferSent` already reliably indicates the request reached the server

## Details

During handshake reconciliation, the BLoC previously skipped **all** unacknowledged outgoing calls, even those whose `OutgoingCallRequest` had silently failed. Such calls were never in the server's line list, so they would be skipped on every subsequent handshake and remain stuck forever.

The fix: skip only calls where `processingStatus` is earlier than `outgoingOfferSent` (i.e. the request was never sent). If the status is `outgoingOfferSent` or `outgoingRinging` but the server has no record of the call → force-terminate it as dead.

### DIFF
--- a/lib/features/call/bloc/call_bloc.dart
+++ b/lib/features/call/bloc/call_bloc.dart
@@ -2563,13 +2563,11 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
       }
 
       // Skips outgoing calls that are still in-flight (OutgoingCallRequest not yet sent).
-      // Once processingStatus reaches outgoingOfferSent or later, the request has reached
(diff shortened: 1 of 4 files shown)


digiboridev (approved):

SERDUN:
Closing in favour of a fresh branch rebased on current develop.

PR #1077, opened 2026-04-07
refactor: move session buffer from SignalingHubModule to SignalingHubClient

SERDUN:
## Summary

Eliminates the duplicate `SignalingEventBuffer` in `SignalingHubModule` by moving it one level down into `SignalingHubClient`.

**Before:** two independent buffers in the pipeline:
- `SignalingHub._sessionBuffer` -- replay for new hub subscribers (cross-isolate)
- `SignalingHubModule._eventBuffer` -- replay for late `hubModule.events` subscribers

**After:** one buffer per pipeline level:
- `SignalingHub._sessionBuffer` -- replay for new hub subscribers (cross-isolate, unchanged)
- `SignalingHubClient._eventBuffer` + `snapshot` -- accumulates decoded events; `SignalingHubModule.events` replays from `_hubClient.snapshot` instead of maintaining its own buffer

`SignalingHubModule` no longer owns any buffering logic -- it delegates replay entirely to the client layer.

## Test plan

- [ ] `dart analyze lib/` -- no issues
- [ ] All 145 tests pass

### DIFF
--- a/packages/webtrit_signaling_service/webtrit_signaling_service_android/lib/src/hub/signaling_hub_client.dart
+++ b/packages/webtrit_signaling_service/webtrit_signaling_service_android/lib/src/hub/signaling_hub_client.dart
@@ -44,11 +44,20 @@ class SignalingHubClient {
   final SendPort _hubPort;
   final ReceivePort _receivePort = ReceivePort();
   final _controller = StreamController<SignalingModuleEvent>.broadcast();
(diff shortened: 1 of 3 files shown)


steamingscut-dev:
Closing: buffer count stays at 2 regardless of where the second buffer lives. Not a meaningful improvement.

PR #1088, opened 2026-04-09
fix: avoid Vulkan/Impeller init in signaling background engine

SERDUN:
## Overview

In `pushBound` mode the signaling WebSocket runs inside an Android foreground service that starts a separate Flutter engine (background isolate). The Dart entry point of that engine — `signalingServiceCallbackDispatcher` — used `WidgetsFlutterBinding.ensureInitialized()` to set up platform channels required by Pigeon.

`WidgetsFlutterBinding` initialises the full rendering stack (`RendererBinding` → Impeller → Vulkan). A background Flutter engine has no rendering surface and does not need any of that — it only needs a `BinaryMessenger` to communicate over platform channels.

## Changes

**Replace binding initialisation:**
`WidgetsFlutterBinding.ensureInitialized()` → `BackgroundIsolateBinaryMessenger.ensureInitialized(token)`, which sets up only a `BinaryMessenger` without touching the rendering stack. `WidgetsFlutterBinding` is retained as a fallback when the token is unexpectedly absent.

**Extract helper and pass messenger explicitly:**
The initialisation is extracted into `_ensureBinaryMessengerInitialized()` which returns `BinaryMessenger?`. The result is passed explicitly to both `PSignalingServiceFlutterApi.setUp(binaryMessenger: ...)` and `PSignalingServiceHostApi(binaryMessenger: ...)`, removing the implicit dependency on `ServicesBinding.defaultBinaryMessenger`.

The helper name reflects what it actually does — sets up a `BinaryMessenger`, not a Flutter binding.

## Files changed

- `packages/webtrit_signaling_service/webtrit_signaling_service_android/lib/src/isolate/entry_point.dart`

## Test plan

- [ ] Build and install a debug APK
- [ ] Log in, make a call to confirm signaling works normally
- [ ] Confirm logcat shows `BackgroundIsolateBinaryMessenger initialised` from the background isolate thread (no Impeller/Vulkan init)
- [ ] Confirm `synchronizeIsolate succeeded` appears in logcat within 1–2 s of service start

### DIFF
--- a/packages/webtrit_signaling_service/webtrit_signaling_service_android/lib/src/isolate/entry_point.dart
+++ b/packages/webtrit_signaling_service/webtrit_signaling_service_android/lib/src/isolate/entry_point.dart
@@ -1,6 +1,6 @@
 import 'dart:ui' show RootIsolateToken;
 
-import 'package:flutter/services.dart' show BackgroundIsolateBinaryMessenger;
+import 'package:flutter/services.dart' show BackgroundIsolateBinaryMessenger, BinaryMessenger;
 import 'package:flutter/widgets.dart' show WidgetsFlutterBinding;
 import 'package:logging/logging.dart';
 import 'package:logging_appenders/logging_appenders.dart';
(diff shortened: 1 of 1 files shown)


SERDUN:
Closing — the BackgroundIsolateBinaryMessenger approach is incompatible with bidirectional Pigeon channels (setMessageHandler is not supported). The Samsung Vulkan hang requires a different fix at the Kotlin level.

PR #1090, opened 2026-04-10
fix(android): minimize app instead of destroying Activity when back pressed at root

SERDUN:
## Problem

When a user presses Back on the root screen while an active call is in progress, Android destroys `MainActivity` instead of moving it to the background. This tears down the Flutter engine and the entire WebRTC stack, causing the call to drop.

Observed sequence in logcat:
1. `onKeyDown(KEYCODE_BACK)` on root screen
2. `DestroyActivityItem.execute` + `FlutterActivity.onDestroy` — Activity destroyed
3. `AppLifecycleState.detached` — Flutter engine gone
4. `ice_hangup` with `reason: DTLS alert` — WebRTC torn down
5. Server sends `hangup` with `reason: to BYE` — call ended

## Fix

Override `popSystemNavigator()` in `MainActivity`. This method is called by Flutter **only after its own navigation stack is fully exhausted** — meaning Flutter has already handled all internal back presses. At that point, instead of the default `finish()`, we call `moveTaskToBack(true)` to minimize the app.

```kotlin
override fun popSystemNavigator(): Boolean {
    moveTaskToBack(true)
    return true
}
```

## Behavior

| Situation | Result |
|---|---|
| Flutter has routes in stack | Flutter pops route normally, `popSystemNavigator` not called |
| Flutter stack is empty (root screen) | `moveTaskToBack(true)` — app minimizes, engine stays alive |

### DIFF
--- a/android/app/src/main/kotlin/com/webtrit/app/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/webtrit/app/MainActivity.kt
@@ -3,11 +3,12 @@ package com.webtrit.app
 import io.flutter.embedding.android.FlutterActivity
 
 class MainActivity : FlutterActivity() {
-    // Move the app to background instead of destroying the Activity when
-    // the user presses Back on the root screen. This prevents the Flutter
-    // engine (and any active WebRTC call) from being torn down on back press.
-    @Suppress("DEPRECATION")
-    override fun onBackPressed() {
+    // Called by Flutter only after its own navigation stack is fully exhausted.
+    // Returning true prevents the default finish() call and moves the app to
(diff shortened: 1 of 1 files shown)


SERDUN:
Addressed both review comments:

1. **moveTaskToBack fallback** — now returns `super.popSystemNavigator()` if `moveTaskToBack` returns false (e.g. activity is not a root task).

2. **Always-minimize scope** — kept intentional. VoIP apps must stay alive in background for signaling and incoming calls, so minimizing on back at root is the correct default. Added KDoc explaining this explicitly. Users can still fully exit via the recent-apps screen.

digiboridev (approved):

PR #1162, opened 2026-04-19
fix(hub): replay last registration event to late subscribers

SERDUN:
## Problem

On devices without Google services (Huawei, persistent mode only), reopening the app after it was backgrounded causes the registration status to get stuck at **Unregistered**.

## Root cause

The initial `StateHandshake` from the server always carries `registration.status = unregistered` because SIP REGISTER has not completed yet when the WebSocket handshake is sent. Once registration succeeds, the server sends a `RegisteredEvent` — a `SessionEvent` wrapped as `SignalingProtocolEvent`.

`SignalingProtocolEvent`s are intentionally excluded from `_sessionBuffer` (to avoid unbounded growth from call events). This means that when the main app opens and subscribes to the hub, it only receives the buffered `[Connecting, Connected, Handshake(unregistered)]` replay and never sees `RegisteredEvent`. Result: stuck at `Unregistered`.

## Fix

Track the last registration-state protocol event (`RegisteredEvent`, `UnregisteredEvent`, `RegisteringEvent`, `RegistrationFailedEvent`, `UnregisteringEvent`) in a separate `_lastRegistrationEvent` field. Replay it to new subscribers after the session buffer, so they always receive the current registration state regardless of when they subscribe.

The field is cleared on each `SignalingConnecting` (new WebSocket session) and overwritten on each state transition — only the most recent registration event is kept.

## Affected scenario

- Device without Google services (Huawei, Xiaomi without GMS, etc.)
- Persistent connection mode (FGS stays alive in background)
- App reopened after being backgrounded for any duration
- First install works fine (app subscribes before `RegisteredEvent` arrives)
- Subsequent opens broken (app subscribes after `RegisteredEvent` already fired)

### DIFF
--- a/packages/webtrit_signaling_service/webtrit_signaling_service_android/lib/src/hub/signaling_hub.dart
+++ b/packages/webtrit_signaling_service/webtrit_signaling_service_android/lib/src/hub/signaling_hub.dart
@@ -63,6 +63,19 @@ class SignalingHub {
   /// and would cause the buffer to grow unboundedly over a long session.
   final List<List<dynamic>> _sessionBuffer = [];
 
+  /// Last encoded registration-state protocol event in the current session.
+  ///
+  /// The initial [StateHandshake] always carries [RegistrationStatus.unregistered]
+  /// because SIP REGISTER hasn't completed yet. The server then sends a
+  /// [RegisteredEvent] (or similar) as a protocol event, which is normally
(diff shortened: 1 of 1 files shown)


SERDUN:
Merged into #1160 (fix/fgs-callkeep-incoming-call-persistent) via cherry-pick.

PR #1170, opened 2026-04-20
fix(signaling): self-healing timer to recover from stuck _startPending

SERDUN:
## What and why

`WebtritSignalingService.connect()` has a guard flag `_startPending` that is set to `true` on entry and cleared only when a terminal event (`SignalingConnected` / `SignalingDisconnected` / `SignalingConnectionFailed`) arrives from the background isolate.

Two failure modes can prevent that event from ever arriving:

1. **TCP-level hang** — the background isolate starts a WebSocket upgrade whose TCP connect stalls (e.g. a middlebox silently drops SYN packets). Android's default TCP timeout can exceed 10 minutes, so no OS error is delivered and the isolate never emits a failure event.

2. **Background-isolate state mismatch** — after the app resumes from background, `SignalingReconnectController` calls `connect()` via a one-shot timer. If `_startPending` is already `true` from a prior stalled attempt, the call is silently discarded. Because `SignalingReconnectController` has no periodic timer (it reschedules only on terminal events or external lifecycle notifications), the reconnect loop goes permanently silent until the next user-visible lifecycle change.

**Fix:** a `_startPendingTimer` is armed on every `connect()` call. If `kSignalingStartPendingTimeout` (30 s) elapses with no terminal event, `_startPending` is reset and `connect()` is retried automatically. The timer is cancelled immediately when any terminal event arrives, so it has no effect during normal operation.

Observed on Samsung SM-A505FN (Android 11): `notifyAppResumed` triggered `connect()` three times, all silently skipped — no WebSocket attempt was ever made.

## Changes

- `lib/app/constants.dart` — added `kSignalingStartPendingTimeout = Duration(seconds: 30)`
- `packages/webtrit_signaling_service/.../signaling_service.dart` — self-healing timer, `_clearStartPending()` helper, full docblock on `connect()`
- `lib/app/router/main_shell.dart` — passes `startPendingTimeout: kSignalingStartPendingTimeout`

## Test to reproduce before merge

> These steps simulate a TCP-level hang so `_startPending` never gets cleared by a normal terminal event.

1. On the server, add a firewall rule that **DROPs** (not rejects) packets on the WebSocket port:
   ```bash
   iptables -I INPUT -p tcp --dport <ws_port> -j DROP
   ```
2. Build with `kSignalingStartPendingTimeout = Duration(seconds: 5)` for a faster feedback loop.
3. Launch the app — it connects normally. Put it in background.
4. Activate the DROP rule so the next reconnect attempt hangs.
5. Resume the app (or trigger `notifyAppResumed`).
6. Observe in logcat:
   ```
   connect: no terminal event after 0:00:05.000000 — resetting and retrying
   ```
7. Remove the DROP rule. Within the next 5 s the app should reconnect and reach `SignalingConnected`.
8. Restore `kSignalingStartPendingTimeout = Duration(seconds: 30)` before merging.

### DIFF
--- a/lib/app/constants.dart
+++ b/lib/app/constants.dart
@@ -15,6 +15,13 @@ const kCallRoutingStateTimeout = Duration(seconds: 10);
 const kSignalingClientReconnectDelay = Duration(seconds: 3);
 const kSignalingClientFastReconnectDelay = Duration(seconds: 1);
 
+/// How long [WebtritSignalingService.connect] waits for a terminal event
+/// (Connected / Disconnected / ConnectionFailed) before resetting
+/// [_startPending] and retrying. Covers the case where the background isolate
+/// accepts the connect command but never sends a response — e.g. a WebSocket
+/// upgrade that hangs at the OS/TCP level without timing out.
+const kSignalingStartPendingTimeout = Duration(seconds: 30);
+
 const kPeerConnectionRetrieveTimeout = Duration(seconds: 5);
 
 const kCompatibilityVerifyRepeatDelay = Duration(seconds: 2);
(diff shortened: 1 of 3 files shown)


SERDUN:
@claude Please review this PR

PR #1206, opened 2026-04-23
fix(signaling): fix 8-hour reconnect loop silence on Android FGS (WT-1403)

SERDUN:
## Summary

Root cause analysis of WT-1403 (Xiaomi/pixel devices stuck without signaling for hours after background) revealed a **two-bug chain** in `SignalingModuleImpl` plus a **missing error type** in the cross-isolate codec. Three independent fixes:

**Bug #1 — `_errorHandled` not reset on reconnect** (`85110c822`)
When `_onError` fires for clientA it sets `_errorHandled = true`. If clientB then connects without resetting this flag, clientB's legitimate disconnect is silently swallowed in `_onDisconnect` — no `SignalingDisconnected` event is emitted, `SignalingReconnectController` never fires, loop goes silent.
Fix: `_errorHandled = false` in `_connectAsync` after `_client = client`.

**Bug #2 — stale callback guard hole when `_client == null`** (`c6ffcf848`)
Old guard: `_client != null && !identical(_client, client)`. When `_client` is already null (cleared by `_onError`), any stale `_wsOnDone` from a superseded client passes through, emitting a spurious `SignalingDisconnected(delay=null)`. `SignalingReconnectController` interprets `delay=null` as "do not reconnect" and breaks its loop.
Fix: `_activeClientId` — unique `Object` token per client, captured in closures before `client.listen()`, cleared in both `_onError` and `_onDisconnect`. Guard: `!identical(_activeClientId, myId)`.

**Bug #3 — ghost-state detection via `NotConnectedException`** (`c6ffcf848`)
When the FGS hub is disconnected but main-isolate `_isConnected = true`, `SignalingHub._executeAndReply` threw `StateError` which the codec serialized as a generic `Exception`. `WebtritSignalingService.execute()` couldn't identify it and skipped recovery. Now the hub throws `NotConnectedException`, the codec encodes/decodes it by type, and `execute()` detects it via `catchError` → resets `_isConnected = false` → calls `connect()`.

## Test plan

- [ ] 8 new tests in `signaling_module_impl_test.dart` (4 BUG-reproducing + 4 after-fix) for Bug #1 and #2
- [ ] 5 new tests in `signaling_service_test.dart` for Bug #3 (ghost-state reset, reconnect trigger, rethrow, normal path, non-NCE path)
- [ ] All 85 tests pass (`flutter test` in platform_interface + service packages)
- [ ] Reproduce on Xiaomi device: connect → background 8+ hours → foreground → verify reconnect within seconds

### DIFF
--- a/packages/webtrit_signaling_service/webtrit_signaling_service/lib/src/signaling_service.dart
+++ b/packages/webtrit_signaling_service/webtrit_signaling_service/lib/src/signaling_service.dart
@@ -166,11 +166,20 @@ class WebtritSignalingService implements SignalingModule {
   @override
   Future<void>? execute(Request request) {
     if (_isConnected) {
-      return _requestQueue.executeNow(
-        execute: SignalingServicePlatform.instance.execute,
-        request: request,
-        isActive: () => _isConnected,
-      );
+      return _requestQueue
+          .executeNow(
+            execute: SignalingServicePlatform.instance.execute,
+            request: request,
+            isActive: () => _isConnected,
+          )
+          ?.catchError((Object e, StackTrace s) {
+            if (e is NotConnectedException) {
+              _logger.warning('execute: ghost state detected — forcing reconnect', e, s);
+              _isConnected = false;
+              connect();
+            }
+            Error.throwWithStackTrace(e, s);
+          });
     }
     return _requestQueue.enqueue(request);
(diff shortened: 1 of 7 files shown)


SERDUN:
@claude Please review this PR

PR #1250, opened 2026-05-04
refactor: replace SessionStatus enum with class (signalingStatus + pushTokenError)

SERDUN:
## Summary

- `SessionStatus` is now a value class instead of an enum
- Holds two independent fields: `signalingStatus: CallStatus` (signaling layer) and `pushTokenError: String?` (push token layer)
- `pushTokenError` was semantically orthogonal to signaling status — merging them into one enum created an awkward special case in cubit mapping and blocked clean use of `isTransientReconnecting`
- Implements `==` and `hashCode` so Freezed equality works correctly
- All extensions (`l10n`, `appBarl10n`, `color`, `key`) updated: guard `if (hasPushTokenError)` first, then `switch (signalingStatus)`
- `MainAppBar` comparisons migrated to `.signalingStatus` and `.hasPushTokenError`
- `patrol_test` updated to use new constructor API

## Related

Part of the WT-1431 refactor plan — prepares ground for moving debounce logic to UI layer (toolbar).

### DIFF
--- a/lib/extensions/session_status.dart
+++ b/lib/extensions/session_status.dart
@@ -35,8 +35,7 @@ extension SessionStatusColor on SessionStatus {
     final colorScheme = themeData.colorScheme;
     final callStatusStyles = themeData.extension<CallStatusStyles>()?.primary;
 
-    // TODO(Serdun): Move pushTokenError color to the color scheme
-    if (hasPushTokenError) return Colors.orange;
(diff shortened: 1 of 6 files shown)


SERDUN:
@claude Please review this PR.

digiboridev (approved):

PR #1275, opened 2026-05-19
fix: nsn matching only for nsn part (WT-1535)

digiboridev:
Improves contact info builder phone matching to match nsn part only for nsn parsed numbers.

### DIFF
--- a/lib/repositories/contacts/contacts_repository.dart
+++ b/lib/repositories/contacts/contacts_repository.dart
@@ -204,44 +204,38 @@ class ContactsRepository
     );
   }
 
-  Stream<Contact?> watchContactByPhoneNumber(String number) {
-    return _appDatabase.contactsDao.watchContactByPhoneNumber(number).map((data) {
-      if (data == null) return null;
-      return contactFromDrift(
-        data.contact,
-        phones: data.phones,
-        emails: data.emails,
-        favorites: data.favorites,
-        presenceInfo: data.presenceInfo,
-        dialogInfo: data.dialogInfo,
-      );
-    });
-  }
+  Stream<Contact?> watchContactByPhoneNumber(String number) async* {
+    await for (var data in _appDatabase.contactsDao.watchContactByPhoneNumber(number)) {
+      // In case if exact phone number is not found try to find by national significant number (NSN)
+      //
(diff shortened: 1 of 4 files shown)


lib/repositories/contacts/contacts_repository.dart
```diff
@@ -204,44 +204,38 @@ class ContactsRepository
...
+      // and CDR record becomes `+380507259336`, but local contact number stays `0507259336`.
+      // Case 2: if user just called using full international number `+380507259336` from keypad but local contact saved as `0507259336`.
+      //
+      // In these cases we need to find contact by national significant number `0507259336`.
+      // To do it we verify if that number is truly valid full number and only than extract NSN part.
+      //
+      // Also check length to reduce libphonenumbers evaluation
+      if (data == null && number.length >= 6) {
+        final nsnPart = number.nationalPhoneIfValid;
+        if (nsnPart != null && nsnPart != number) {
```
SERDUN: Looks like the check in final if (nsnPart != null && **nsnPart != number**) is redundant - it's already done in nationalPhoneIfValid.

lib/repositories/contacts/contacts_repository.dart
```diff
@@ -204,44 +204,38 @@ class ContactsRepository
...
+      // Case 1: if user has local contact like '0507259336', but after call PBX automatically appends country code `+380`
+      // and CDR record becomes `+380507259336`, but local contact number stays `0507259336`.
+      // Case 2: if user just called using full international number `+380507259336` from keypad but local contact saved as `0507259336`.
+      //
+      // In these cases we need to find contact by national significant number `0507259336`.
+      // To do it we verify if that number is truly valid full number and only than extract NSN part.
+      //
+      // Also check length to reduce libphonenumbers evaluation
+      if (data == null && number.length >= 6) {
+        final nsnPart = number.nationalPhoneIfValid;
```
SERDUN: Maybe it might be worth moving `final nsnPart = number.nationalPhoneIfValid;` outside of `await for (var data in _appDatabase.contactsDao.watchContactByPhoneNumber(number))` - since `number` appears to be immutable, we could compute it just once.

SERDUN (approved):

PR #1280, opened 2026-05-20
fix: call drop on network interface switch (WiFi <-> LTE)

SERDUN:
## Problem

On iOS, when the network interface switches (WiFi → LTE or back), active calls were being dropped. Root cause: the existing TCP/WebSocket connection becomes a **zombie** — appears alive at the socket level but packets are no longer delivered on the new interface. Signaling transactions exhaust their retries and kill the call even though the ICE layer recovers on the new interface.

### Fix D — Proactively kill zombie WebSocket on interface change (root fix)
When `notifyNetworkAvailable()` fires (connectivity plugin detects interface change), explicitly disconnect the stale WebSocket before scheduling a reconnect. Forces a fresh connection on the new interface within ~1–3s instead of waiting for the ping timeout (~20s).

`notifyNetworkAvailable()` is now `Future<void>` so callers can await the disconnect. On Android, where the OS closes stale sockets on interface change, `isConnected` is already false and the call is a no-op.

### Fix C — WebSocket protocol-level ping as fallback
Sets `pingInterval = 10s` on the underlying `io.WebSocket`. The Dart IO layer sends a binary ping frame every 10s and closes the socket (code 1001) if no pong is received before the next ping fires (~20s total). Catches the case where the OS does not emit a network-change event.

`wsPingInterval` exposed as a constructor parameter on `SignalingModuleImpl` (default: `WebtritSignalingClient.defaultWsPingInterval = 10s`) and threaded through `SignalingClientFactory` for test injection.

### Fix B — SDP rollback instead of call termination on signaling disconnect during renegotiation (resilience layer)
During ICE restart renegotiation, the WebSocket may disconnect before the offer transaction completes. Previously this terminated the call via `callErrorReporter`. Now, when `WebtritSignalingTransactionTerminateByDisconnectException` is caught and the PeerConnection state is not Failed/Closed, the local SDP offer is rolled back via `setLocalDescription('', 'rollback')`. This returns the PC to Stable state so that when signaling reconnects and triggers `onRenegotiationNeeded` again, a fresh offer with updated ICE credentials is generated and sent successfully.

## Recovery flow

```
WiFi → LTE switch
  ↓
notifyNetworkAvailable() [Fix D]
  → await module.disconnect() (kills zombie TCP immediately)
  → scheduleReconnect(kSignalingClientFastReconnectDelay)
  → new WS on LTE interface (~1-3s)
  
[If Fix D missed] pingInterval=10s [Fix C]
  → WS closed after ~20s with code 1001
  → SignalingReconnectController schedules reconnect
  
[If signaling disconnected mid-renegotiation] [Fix B]
  → catch WebtritSignalingTransactionTerminateByDisconnectException
  → rollback setLocalDescription
  → wait for signaling reconnect → re-trigger onRenegotiationNeeded
  → fresh ICE restart offer sent
```

## Test plan
- [ ] iOS: make a call, switch WiFi → LTE → call stays active
- [ ] iOS: make a call, switch LTE → WiFi → call stays active
- [ ] Android: regression — call behaviour unchanged on interface switch
- [ ] All unit tests pass (`flutter test`)

### DIFF
--- a/lib/features/call/bloc/call_bloc.dart
+++ b/lib/features/call/bloc/call_bloc.dart
@@ -516,7 +516,7 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
     if (connectivityState == ConnectivityResult.none) {
       _reconnectController.notifyNetworkUnavailable();
     } else {
-      await _reconnectController.notifyNetworkAvailable();
+      _reconnectController.notifyNetworkAvailable();
     }
 
     WebRTC.initialize(options: webRtcOptionsBuilder?.build());
@@ -541,7 +541,7 @@ class CallBloc extends Bloc<CallEvent, CallState> with WidgetsBindingObserver im
     if (connectivityResult == ConnectivityResult.none) {
       _reconnectController.notifyNetworkUnavailable();
     } else {
-      await _reconnectController.notifyNetworkAvailable();
+      _reconnectController.notifyNetworkAvailable();
 
       // Restart ICE for all active calls to trigger faster recovery from connectivity loss.
       //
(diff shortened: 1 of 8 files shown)


SERDUN:
@claude Please review my PR.

PR #1318, opened 2026-06-02
fix(signaling): stop SignalingForegroundService crash loop on boot (WT-1524)

SERDUN:
> ⚠️ **This is a hotfix** (target **mob/1.15.3**) — it stops the production crash loop now. The full architectural solution will be implemented in one of the upcoming tasks (see *Long-term fix* below).

## WT-1524 — SignalingForegroundService crashes on boot (`phoneCall` FGS not allowed from `BOOT_COMPLETED`)

YouTrack: https://youtrack.portaone.com/issue/WT-1524 · Target: **mob/1.15.3** · Severity: Medium

### Problem
On **Android 15+**, after a device reboot with persistent mode active, `SignalingForegroundService` crashed:

```
ForegroundServiceStartNotAllowedException: FGS type phoneCall not allowed to start from BOOT_COMPLETED!
  at SignalingForegroundService.startForeground(...)
  at SignalingForegroundService.onCreate(...)
```

`SignalingBootReceiver` started the service from `BOOT_COMPLETED`; `startForeground()` threw inside `onCreate()` on the main thread (after `startForegroundService()` had already returned on another thread), and `START_STICKY` restarts inherited the same boot context → **user-visible crash loop**. Regression from PR #1244 (`remoteMessaging` → `phoneCall` FGS type).

### Hotfix (this PR)
1. **Remove `SignalingBootReceiver`** + its `BOOT_COMPLETED`/`LOCKED_BOOT_COMPLETED`/`MY_PACKAGE_REPLACED` intent-filter, and the now-unused `RECEIVE_BOOT_COMPLETED` permission from the **signaling module** manifest.
2. **Guard `ServiceCompat.startForeground()`** in try-catch: on `ForegroundServiceStartNotAllowedException` (API 31+) `stopSelf()` instead of crashing. `onDestroy` already re-enqueues `SignalingRestartWorker` (delay 15 s), so the crash loop degrades to a slow retry. This is the only place on the correct thread that can stop the crash, and it also covers `START_STICKY` re-creations under memory pressure — which removing the boot receiver alone does not.
3. **Reminder dialog** (`persistentConnectionReminder*`, en/uk/it) already shown on persistent selection in `network_screen.dart` — **unchanged** (only confirmed it still triggers).

> The app-level `RECEIVE_BOOT_COMPLETED` (under the *push notifications* group in `android/app/.../AndroidManifest.xml`) is left untouched — it serves push, not signaling.

### Behaviour after this hotfix
- **No autostart of persistent mode on any platform.** Removing the boot receiver means persistent mode no longer comes up on its own after a reboot / app update. It resumes only after the user opens the app at least once.
- The user is already **warned about this in the UI**: selecting this incoming-call type shows the reminder dialog (`persistentConnectionReminder*`) explaining they must launch the app manually after a restart to re-establish the persistent connection.
- This is an intentional trade-off for the hotfix: a `phoneCall` FGS cannot be started from any background context by design, so stopping the crash necessarily disables boot autostart for now.

### Long-term fix (separate upcoming task)
The proper solution — restoring autostart without the crash — will be implemented in one of the next tasks: **split FGS types** (`specialUse` listener allowed from `BOOT_COMPLETED` + `phoneCall` for the active call, per the Linphone pattern), with a `<property>` declaration and Play Console justification. Not part of this hotfix.


SERDUN:
Superseded by #1319 — branch renamed `hotfix/…` → `fix/…` to satisfy the branch-name lint (allowed prefixes: feat|feature|refactor|fix|chore|build|style|docs|release). Same commit, same changes.

PR #1355, opened 2026-06-08
fix: prefer backend contact over local phonebook on number collision

SERDUN:
## Problem

When a number exists both as a local phonebook contact and as an external/PBX
extension, the call screen could show the local name instead of the PBX alias.
The lookup `getContactByPhoneNumber` used `LIMIT 1` with no `ORDER BY`, so the
winner was decided by SQLite's unspecified row order (platform/data dependent) -
latent and non-deterministic (it surfaced on 1.13.1; does not reproduce on host).

## Fix

Order the phone-number lookups by `sourceType` DESC so the external (PBX) entry
(`sourceType` index 1) wins deterministically over local (index 0). Applied to
`getContactByPhoneNumber`, `watchContactByPhoneNumber` and both
`PhoneMatchedEnding` variants.

## Tests

`contacts_dao_test.dart`: local-vs-external number collision in BOTH insertion
orders -> external/PBX wins (locks the ordering contract).

## Verification

- [x] `dart format` + `dart analyze` clean
- [x] full `flutter test` suite green (pre-push hook)
- [ ] Manual: number collision -> PBX alias shown

Ref: WT-1037
Related: #1356 (own alias on self-call)

### DIFF
--- a/packages/data/app_database/lib/src/daos/contacts_dao.dart
+++ b/packages/data/app_database/lib/src/daos/contacts_dao.dart
@@ -174,6 +174,7 @@ class ContactsDao extends DatabaseAccessor<AppDatabase> with _$ContactsDaoMixin 
   Future<FullContactData?> getContactByPhoneNumber(String number) async {
     final query = _joinFullData(select(contactsTable))
       ..where(contactPhonesTable.number.equals(number))
+      ..orderBy([OrderingTerm(expression: contactsTable.sourceType, mode: OrderingMode.desc)])
       ..limit(1);
 
     return query.get().then(_gatherSingleContact);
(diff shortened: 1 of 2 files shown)


SERDUN:
Addressed the review feedback (commit ef8affbb):

- Collision regression coverage: the test now parametrizes the collision (a local
  and an external contact sharing the same number) over EVERY lookup the ORDER BY
  fix touches - getContactByPhoneNumber, watchContactByPhoneNumber, and both
  PhoneMatchedEnding variants - in both insertion orders, asserting the external
  (PBX) contact wins. 8 cases total.
- Renamed the local test contact "Loca Contact 4" -> "Local Contact 4" (setup and
  assertions kept consistent).

PR #1398, opened 2026-06-12
fix: localize missed call notification title and unknown caller fallback

SERDUN:
## Overview

Missed call local push notifications always showed "Missed Call" and "Unknown" in English, regardless of the device locale. This was a hardcoded issue in both the background push isolate and the main `CallBloc`.

- `AppLocalPush.missedCall()` - removed hardcoded `'Missed Call'` title and `TODO` comment; factory now accepts `title` as an explicit param
- `PushNotificationIsolateManager` - added `missedCallTitle` / `unknownCallerFallback` constructor params; `PushIsolateContext` now reads the persisted locale via `LocaleRepositoryPrefsImpl` and exposes it as `locale`; `background_isolate_callbacks.dart` calls `lookupAppLocalizations(context.locale)` (no BuildContext needed) to resolve strings before constructing the manager
- `CallBloc` - added `missedCallTitle` constructor param; passed from `main_shell.dart` via `context.l10n`
- New ARB keys `notifications_missedCall_title` / `notifications_missedCall_unknownCaller` added to EN, IT, UK

Fixes WT-1088.


SERDUN:
Merged.

PR #1399, opened 2026-06-12
fix: enforce exclusive voicemail playback via shared AudioPlayer (WT-1363)

SERDUN:
## Overview

Multiple voicemails could play simultaneously because each `AudioView` tile
owned its own `AudioPlayer` instance with no cross-tile coordination.

Introduces `VoicemailPlaybackController` -- a `ChangeNotifier` that holds a
single `AudioPlayer` for the entire voicemail screen. Switching to a new
track stops the previous one automatically since only one player exists.

## Changes

- **`VoicemailPlaybackController`** (new) -- `ChangeNotifier` + `WidgetsBindingObserver`
  that owns one `AudioPlayer`. Handles audio session setup,
  `LockCachingAudioSource` construction, playback-completed reset, and
  app-lifecycle pause. Provided via `ChangeNotifierProvider` in
  `VoicemailScreenPage`.
- **`AudioView`** -- refactored from `StatefulWidget` to `StatelessWidget`.
  Reads the controller from context. The active tile (matching `activeId`)
  renders `AudioPlayerInterface` against the shared player; all other tiles
  render a static `_InactiveAudioView` play button. Loading and error states
  are also driven by the controller.
- **`VoicemailScreenContext`** -- unchanged; remains a plain config bag.

## Test plan

- [ ] Open Voicemail screen with multiple messages
- [ ] Tap Play on first message -- it plays
- [ ] Without stopping, tap Play on second message -- first stops, second plays
- [ ] Tap pause on active message -- pauses correctly
- [ ] Tap play again -- resumes from paused position
- [ ] Let a message play to end -- player resets to beginning
- [ ] Background the app while playing -- playback stops
- [ ] Return to foreground -- play button shown correctly
- [ ] Retry after a load error works

WT-1363

### DIFF
--- a/lib/features/settings/features/voicemail/bloc/bloc.dart
+++ b/lib/features/settings/features/voicemail/bloc/bloc.dart
@@ -1,1 +1,2 @@
 export 'voicemail_cubit.dart';
+export 'voicemail_playback_controller.dart';
--- a/lib/features/settings/features/voicemail/bloc/voicemail_playback_controller.dart
+++ b/lib/features/settings/features/voicemail/bloc/voicemail_playback_controller.dart
@@ -24,6 +24,7 @@ class VoicemailPlaybackController extends ChangeNotifier with WidgetsBindingObse
   String? _activeId;
   bool _isLoading = false;
   Object? _error;
+  Timer? _loadingDebounce;
 
   String? get activeId => _activeId;
   bool get isLoading => _isLoading;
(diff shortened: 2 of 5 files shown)


SERDUN:
## Overview

Addressed all Copilot review findings in `7033a7d4`:

- **Retry after load error** (`voicemail_playback_controller.dart`): `play()` now falls through its same-id fast-path when `_error != null`, so retrying an errored tile re-runs `setAudioSource` instead of resuming an unloaded source.
- **Out-of-order async / no stop on error**: added a `_generation` token; every continuation after an `await` bails out if a newer `play()` superseded it, and the `catch` block now calls `player.stop()` so a previously playing voicemail does not keep playing after a failed load.
- **Cache-key path traversal** (`resolveCacheFile`): besides stripping path separators, a `cacheKey` of `.` or `..` is now rejected (returns `null`), since `path.join(base, '..')` would otherwise escape the cache directory.
- **Inactive slider** (`audio_view.dart`): the inactive tile's `AudioSlider` is wrapped in `IgnorePointer`, so the no-op seek control is no longer draggable.

Also added unit tests (incl. race-guard, error recovery, and path-traversal cases); `resolveCacheFile` is exposed via `@visibleForTesting`. Local `flutter analyze` + full `flutter test` pass.

PR #1411, opened 2026-06-19
feat(permissions): guide Xiaomi/HyperOS users to enable lock-screen calls (WT-1349)

SERDUN:
## Problem

On Xiaomi/HyperOS the incoming-call screen does not cover the lock screen, even with all standard permissions granted. The blocker is the OEM capability **"Display pop-up windows while running in background"** (`OP_BACKGROUND_START_ACTIVITY`), which is separate from the user-facing "Show on lock screen" toggle and has no public API. See WT-1349.

## Change

Drives the existing manufacturer tip from the new callkeep `backgroundActivityStart` status:

- The Xiaomi guidance is surfaced only while the capability is actually missing (read via `appPermissions.backgroundActivityStartStatus()`), so it clears itself once the user enables it and returns to the app (re-checked on `resumed`).
- The action button now deep links to the OEM "Other permissions" editor via `openBackgroundActivityStartSettings()` instead of generic app settings.
- Xiaomi instructions updated to point at the background pop-up / lock-screen toggles (en/uk/it).

Non-blocking: this uses the dismissible manufacturer-tip path, so it never traps onboarding. No effect on non-Xiaomi devices.

Depends on webtrit_callkeep PR #337 (consumed via the local path dependency on develop).

WT-1349


lib/features/permissions/cubit/permissions_cubit.dart
```diff
@@ -94,15 +100,22 @@ class PermissionsCubit extends Cubit<PermissionsState> {
...
-    // Determine if we need to set or keep the manufacturer tip
-    return currentTip ?? (hasManufacturer ? ManufacturerTip(manufacturer: manufacturer) : null);
+  /// Resolves manufacturer-specific guidance for reliable lock-screen calls.
+  ///
+  /// On Xiaomi/HyperOS the incoming-call UI cannot cover the lock screen unless
+  /// the OEM "display pop-up windows while running in background" capability is
+  /// granted, which the standard Android permissions do not cover. The tip is
+  /// surfaced only while that capability is missing, so it clears itself once
+  /// the user enables it and returns to the app. A dismissed tip is preserved.
+  Future<ManufacturerTip?> _resolveManufacturerTip() async {
```
SERDUN, line 115: Device-family detection is split. This path gates on `_checkManufacturer()`, which maps `Build.MANUFACTURER` to `Manufacturer { xiaomi }` (only the literal "xiaomi") and returns `null` early for anything else - so the native `backgroundActivityStartStatus()` is never even queried. callkeep's `isXiaomiFamily()` (which drives that status) also accepts "redmi"/"poco". On a device whose MANUFACTURER reports "redmi"/"poco", callkeep reports the capability denied but this method short-circuits to `null` and shows no tip - on exactly the devices WT-1349 targets. Align the phone-side family list with callkeep, or derive applicability from the callkeep status alone.

lib/data/app_permissions.dart
```diff
@@ -180,10 +190,11 @@ class AppPermissions {
...
   Future<void> toSpecialPermissionsSetting(CallkeepSpecialPermissions permission) async {
     try {
-      if (permission == CallkeepSpecialPermissions.fullScreenIntent) {
-        await _webtritCallkeepPermissions.openFullScreenIntentSettings();
-      } else {
-        await _webtritCallkeepPermissions.openSettings();
+      switch (permission) {
+        case CallkeepSpecialPermissions.fullScreenIntent:
+          await _webtritCallkeepPermissions.openFullScreenIntentSettings();
+        case CallkeepSpecialPermissions.backgroundActivityStart:
```
SERDUN, line 208: `backgroundActivityStart` is never added to `AppPermissions._specialPermissions` (still `[fullScreenIntent]`), so it never enters `missingSpecialPermissions`. This `case`, and the matching arms in `special_permission.dart` and `special_permission_tips.dart`, are therefore unreachable - the capability is surfaced only through the manufacturer-tip path. The enum value ends up modeled twice (special-permission plumbing + manufacturer tip). Either route it through a single path or annotate these arms as unreachable, so a maintainer does not wire/test a screen that never renders.

lib/features/permissions/cubit/permissions_cubit.dart
```diff
@@ -37,7 +37,16 @@ class PermissionsCubit extends Cubit<PermissionsState> {
     final missingSpecialPermissions = await _getDeniedSpecialPermissions();
     _logger.info('Denied special permissions: $missingSpecialPermissions');
 
-    emit(state.copyWith(isPermanentlyDenied: isDenied, missingSpecialPermissions: missingSpecialPermissions));
+    final manufacturerTip = await _resolveManufacturerTip();
+    _logger.info('Manufacturer tip: $manufacturerTip');
+
+    emit(
```
SERDUN, line 47: `checkPermissions()` is fire-and-forget (`void ... async`) and `emit`s after awaiting `isDenied`, `_getDeniedSpecialPermissions`, and `_resolveManufacturerTip` (a new platform-channel round-trip), with no `if (isClosed) return;` guard before `emit` - unlike `initiatePermissionFlow`. If the screen is popped (cubit closed) during those awaits on a `resumed` event, the late `emit` throws "Cannot emit new states after calling close". It also runs on every resume with no debounce/in-flight guard, so two close resume events can finish out of order and the slower one overwrites fresher state (e.g. a just-granted tip reappears). Add the `isClosed` guard and consider de-duping concurrent runs.

digiboridev (approved):

PR #1420, opened 2026-06-22
feat(web): web build foundation - compile + boot to login

SERDUN:
## Overview

PR 1 of the decomposed Flutter web support series, targeting the **`feat/web-support`** integration branch (not develop). Brings the app to a clean `flutter build web` that boots to the **login screen**. Every change is gated on `kIsWeb`; native (Android/iOS) behavior is byte-identical.

Verified: `flutter build web` succeeds; `flutter analyze` clean; full test suite green.

## Scope

- **Build**: modernized `web/index.html` bootstrap (`{{flutter_js}}` + `{{flutter_build_config}}` + `_flutter.loader.load`); `native_audio_restart` conditional-import facade so the flutter_webrtc fork's native-only `restartAudio` is not referenced on web.
- **Bootstrap startup guards**: skip DriftIsolate server, NativeLogForwarder, WorkManager, FCM messaging, local notifications, and callkeep/signaling init on web; guard `FirebaseCrashlytics.log` in the root log listener.
- **Database**: open the drift `WasmDatabase` directly (`IsolateDatabase.openWeb`); ship `web/sqlite3.wasm` (3.3.2) + `web/drift_worker.js` (drift 2.33.0); warn on in-memory fallback.
- **Paths/logging**: `AppPath` returns empty paths on web (path_provider unimplemented); in-memory log repo; `NativeLogForwarder` keeps its `dart:io File` null on web.
- **Remote Config + Firebase Installations**: bound init/`getId` on web (no startup hang); skip the realtime RC stream on web; resilient snapshot reads.
- **call_media_manager**: native audio routing is a no-op on web (browser owns it).

## Not in this PR (follow-ups onto `feat/web-support`)

- Signaling federated web plugin (`webtrit_signaling_service_web`).
- Call/callkeep call-time guards (showOverLockscreen, WebRTC.initialize, dialog_info mapper, push_tokens).
- Permissions web subset.
- Web bundle_id override (`WEBTRIT_APP_WEB_BUNDLE_ID`).

On web, login works; reaching the main shell needs the signaling PR.

### DIFF
--- a/lib/common/app_id_provider.dart
+++ b/lib/common/app_id_provider.dart
@@ -1,6 +1,8 @@
 import 'dart:async';
 import 'dart:math';
 
+import 'package:flutter/foundation.dart';
+
 import 'package:firebase_app_installations/firebase_app_installations.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:logging/logging.dart';
@@ -28,7 +30,11 @@ class FirebaseAppIdProvider implements AppIdProvider {
   @override
   Future<String> getId() async {
     try {
-      final id = await FirebaseInstallations.instance.getId();
+      final future = FirebaseInstallations.instance.getId();
+      // On web a misconfigured Installations promise can never settle, which would
+      // hang bootstrap forever (the catch below never fires). Bound it so we fall
+      // back to a locally generated id instead of blocking startup.
+      final id = await (kIsWeb ? future.timeout(const Duration(seconds: 5)) : future);
       await _saveIdToSharedPreferences(id);
       return id;
     } catch (e) {
@@ -41,11 +47,16 @@ class FirebaseAppIdProvider implements AppIdProvider {
   Stream<String> get onIdChange => _idChangeController.stream;
 
   void _initializeIdListener() {
+    // firebase_app_installations has no reliable web implementation for the
+    // onIdChange stream; skip the subscription on web (getId() already provides
+    // the one-shot id with a bounded fallback). On native, guard onError so a
+    // transient stream error does not escape to the zone.
+    if (kIsWeb) return;
     FirebaseInstallations.instance.onIdChange.listen((id) async {
       await _saveIdToSharedPreferences(id);
       _idChangeController.add(id);
       _logger.info('Firebase ID changed: $id');
-    });
+    }, onError: (Object e, StackTrace s) => _logger.warning('Firebase onIdChange stream error', e, s));
   }
 
   Future<void> _saveIdToSharedPreferences(String id) async {
--- a/lib/common/app_id_provider.dart
+++ b/lib/common/app_id_provider.dart
@@ -47,11 +47,8 @@ class FirebaseAppIdProvider implements AppIdProvider {
   Stream<String> get onIdChange => _idChangeController.stream;
 
   void _initializeIdListener() {
-    // firebase_app_installations has no reliable web implementation for the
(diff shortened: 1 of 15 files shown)


lib/common/app_id_provider.dart
```diff
@@ -28,7 +30,11 @@ class FirebaseAppIdProvider implements AppIdProvider {
   @override
   Future<String> getId() async {
     try {
-      final id = await FirebaseInstallations.instance.getId();
+      final future = FirebaseInstallations.instance.getId();
+      // On web a misconfigured Installations promise can never settle, which would
+      // hang bootstrap forever (the catch below never fires). Bound it so we fall
+      // back to a locally generated id instead of blocking startup.
+      final id = await (kIsWeb ? future.timeout(const Duration(seconds: 5)) : future);
```
SERDUN, line 37: **[code-review] Unstable web device id - CONFIRMED (conditional). FIXED in 0401431a.**

Verified against `firebase_app_installations_web` source + a live web run: `getId()` IS implemented on web and returned a real Firebase FID, so this fallback fires only when `getId()` times out (misconfigured/unreachable Firebase on web). In that path `_generateFallbackId()` always minted a fresh random id and overwrote SharedPreferences -> a different `AppInfo.identifier` on every reload.

Fixed: the fallback now reuses the stored id and generates only on first use.

web/index.html
```diff
@@ -188,22 +177,23 @@
...
   <script>
+    {{flutter_js}}
+    {{flutter_build_config}}
+
     window.addEventListener('load', function () {
       var loadingElement = document.querySelector('#loading');
-      _flutter.loader.loadEntrypoint({
-        serviceWorker: {
-          serviceWorkerVersion: serviceWorkerVersion,
+      _flutter.loader.load({
```
SERDUN, line 185: **[code-review] Service worker dropped - CONFIRMED, but intentionally NOT restored.**

Verified from `flutter.js` source: `load()` calls `loadServiceWorker` only when `serviceWorkerSettings` is passed, so this hand-written `load({onEntrypointLoaded})` does skip SW registration.

However, Flutter`s default service worker is itself deprecated and being removed (flutter/flutter#156910 - "developers should decide ... rather than relying on Flutter`s default implementation"). Re-adding it would reintroduce a deprecated mechanism, so it is left dropped by design. If offline/asset caching is wanted on web, add a custom/standalone service worker (separate decision, not this PR).

lib/common/app_id_provider.dart
```diff
@@ -41,11 +47,16 @@ class FirebaseAppIdProvider implements AppIdProvider {
   Stream<String> get onIdChange => _idChangeController.stream;
 
   void _initializeIdListener() {
+    // firebase_app_installations has no reliable web implementation for the
+    // onIdChange stream; skip the subscription on web (getId() already provides
+    // the one-shot id with a bounded fallback). On native, guard onError so a
+    // transient stream error does not escape to the zone.
+    if (kIsWeb) return;
```
SERDUN: **[code-review] onIdChange skipped on web - CONFIRMED (stronger than first rated). FIXED in 0401431a.**

Verified against `firebase_app_installations_web` source: `onIdChange` IS implemented on web. The original justification comment ("no reliable web implementation") was incorrect, and the `kIsWeb` skip needlessly disabled id-rotation propagation on web.

Fixed: the subscription now runs on all platforms, with the onError guard retained.

PR #1421, opened 2026-06-22
feat(web): web implementation of webtrit_signaling_service

SERDUN:
## Overview

PR 2 of the decomposed Flutter web series. **Stacked on #1420** (base `feat/web-foundation`); review/merge after it.

The federated `webtrit_signaling_service` plugin shipped only android + ios, so on web `SignalingServicePlatform.instance` was never registered and entering the main shell threw `StateError: No SignalingModuleFactory registered`.

## Scope
- New federated package **`webtrit_signaling_service_web`** that delegates to `WebtritSignalingServiceDirect` (same model as iOS): the WebSocket signaling runs in the main isolate (web has no foreground/background service). Registered as the web `default_package`.
- `bootstrap._initCallkeep` now registers the signaling module factory on **every** platform (web/iOS need it); only Android-only background services stay gated behind `kIsWeb || !Platform.isAndroid`.

Verified: `flutter build web` green (web plugin registrant includes `WebtritSignalingServiceWeb`), `flutter analyze` clean. With this, signaling connects on web (incoming/outgoing calls work end-to-end - verified live earlier in the combined branch).

### DIFF
--- a/lib/bootstrap.dart
+++ b/lib/bootstrap.dart
@@ -261,24 +261,23 @@ Future<AppPermissions> _createAppPermissions(
 Future<void> _initCallkeep(FeatureAccess featureAccess) async {
   final logger = Logger('bootstrap');
 
-  if (kIsWeb) {
-    // Callkeep background services and the signaling module factory are not
-    // initialized on web.
-    logger.info('callkeep + signaling init skipped on web');
-    return;
-  }
-
   // Registers the factory used by the signaling service to create a [SignalingModule]
   // instance. Must be a top-level function annotated @pragma('vm:entry-point').
   // iOS: stored in memory, called directly in start(). Android: also persisted to
   // SharedPreferences for deserialization in the background isolate.
+  // Required on every platform - web/iOS run the signaling in the main isolate
+  // (WebtritSignalingServiceDirect) and start() throws without a registered factory.
   try {
     await WebtritSignalingService.setModuleFactory(createSignalingModule);
   } catch (e, s) {
     logger.severe('setModuleFactory failed -- signaling may not work in background isolate', e, s);
   }
 
-  if (!Platform.isAndroid) return;
+  // The remaining callkeep services (Android background push isolate,
+  // persistent-mode call-event handler, SMS triggers) are Android-only; web has
+  // no background isolates and iOS does not use them. kIsWeb short-circuits before
+  // the dart:io Platform check (which is unavailable on web).
+  if (kIsWeb || !Platform.isAndroid) return;
 
   // Registers the top-level callback that the native Android side invokes when a push
   // notification arrives in the background. Bootstraps the push isolate and delegates
--- a/lib/bootstrap.dart
+++ b/lib/bootstrap.dart
@@ -270,7 +270,7 @@ Future<void> _initCallkeep(FeatureAccess featureAccess) async {
   try {
     await WebtritSignalingService.setModuleFactory(createSignalingModule);
   } catch (e, s) {
-    logger.severe('setModuleFactory failed -- signaling may not work in background isolate', e, s);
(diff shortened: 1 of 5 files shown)


lib/bootstrap.dart
```diff
@@ -261,24 +261,23 @@ Future<AppPermissions> _createAppPermissions(
...
   // Registers the factory used by the signaling service to create a [SignalingModule]
   // instance. Must be a top-level function annotated @pragma('vm:entry-point').
   // iOS: stored in memory, called directly in start(). Android: also persisted to
   // SharedPreferences for deserialization in the background isolate.
+  // Required on every platform - web/iOS run the signaling in the main isolate
+  // (WebtritSignalingServiceDirect) and start() throws without a registered factory.
   try {
     await WebtritSignalingService.setModuleFactory(createSignalingModule);
   } catch (e, s) {
     logger.severe('setModuleFactory failed -- signaling may not work in background isolate', e, s);
```
SERDUN: **[code-review] (PLAUSIBLE) Misleading log if the web signaling plugin is not registered.** With the web early-return removed, `setModuleFactory` now resolves `SignalingServicePlatform.instance`, whose getter throws `StateError` if `WebtritSignalingServiceWeb.registerWith` has not run (e.g. a web integration-test harness, or a stale/cached web build predating this new plugin). The try/catch then swallows it and logs "signaling may not work in background isolate" - misleading on web (signaling runs in the main isolate), and signaling is silently dead with no accurate diagnostic. The normal path (main.dart ensureInitialized -> bootstrap) is fine. Consider a web-aware message / not swallowing the StateError.

packages/webtrit_signaling_service/webtrit_signaling_service_web/lib/webtrit_signaling_service_web.dart
```diff
@@ -0,0 +1,16 @@
...
+
+/// Web implementation of [SignalingServicePlatform].
+///
+/// Delegates entirely to [WebtritSignalingServiceDirect] - on web there is no
+/// foreground/background service, so the WebSocket always runs in the main
+/// isolate (the same model iOS uses).
+class WebtritSignalingServiceWeb extends WebtritSignalingServiceDirect {
+  WebtritSignalingServiceWeb._();
+
+  static void registerWith(Registrar registrar) {
```
SERDUN, line 15: **[code-review] (PLAUSIBLE) registerWith is not idempotent (iOS sibling is).** This unconditionally constructs a new `WebtritSignalingServiceWeb` and overwrites `SignalingServicePlatform.instance` on every call. The iOS impl memoizes (`_instance ??= ...`). If the generated web registrant runs `registerWith` more than once (e.g. web hot-restart / re-register), a fresh Direct instance replaces the one already holding the registered module factory + active events controller -> subscribers on the old instance stop receiving events and `start()` throws "No SignalingModuleFactory registered". Low likelihood (runs once at engine init) but worth matching the iOS `_instance ??=` guard.

packages/webtrit_signaling_service/webtrit_signaling_service_web/pubspec.yaml
```diff
@@ -0,0 +1,32 @@
...
+dependencies:
+  flutter:
+    sdk: flutter
+  flutter_web_plugins:
+    sdk: flutter
+  webtrit_signaling_service_platform_interface:
+    path: ../webtrit_signaling_service_platform_interface
+  webtrit_signaling:
+    path: ../../webtrit_signaling
+  logging: ^1.3.0
```
SERDUN: **[code-review] (CONFIRMED) Dead `logging` dependency.** The only source file imports just `flutter_web_plugins` and the platform_interface; `logging` is never used here (copy-pasted from the iOS/Android pubspec). It pins `logging` into this web plugin for nothing. Drop it:

```suggestion
```

PR #1422, opened 2026-06-22
feat(web): request only web-supported permissions on web

SERDUN:
## Overview

PR 3 of the decomposed web series. **Stacked on #1421** (base `feat/web-signaling`).

`permission_handler`'s web implementation supports only a subset of permissions; requesting/querying the others (contacts, sms) throws `UnsupportedError`, which crashed the permissions screen on web.

## Scope
- Restrict the requested set to web-supported permissions (microphone, camera) on web.
- Short-circuit the direct contacts helpers (`isContactPermissionGranted` / `requestContactPermission`) to "not granted" on web.
- Native keeps requesting the full set (behavior byte-identical when `!kIsWeb`).

Verified: `flutter analyze` clean. (Special callkeep permissions like fullScreenIntent are already stubbed by `webtrit_callkeep_web` -> granted, so the permissions route guard does not throw.)

### DIFF
--- a/lib/data/app_permissions.dart
+++ b/lib/data/app_permissions.dart
@@ -1,3 +1,5 @@
+import 'package:flutter/foundation.dart';
+
 import 'package:logging/logging.dart';
 import 'package:permission_handler/permission_handler.dart';
 
@@ -17,6 +19,12 @@ class AppPermissions {
 
   /// The list of all possible permissions that the app may request.
   static const _fullPermissions = [..._requiredPermissions, Permission.camera, Permission.contacts, Permission.sms];
+
+  /// Permissions supported by permission_handler's web implementation. Requesting
+  /// or querying any other permission on web throws UnsupportedError, so on web
+  /// the requested set is restricted to these. (contacts/sms are not available on
+  /// web; camera + microphone cover the call use case.)
+  static const _webSupportedPermissions = [Permission.microphone, Permission.camera];
 
   /// A list of special permissions that are absolutely required for the app to function correctly.
   static const _requiredSpecialPermissions = [CallkeepSpecialPermissions.fullScreenIntent];
@@ -51,10 +59,17 @@ class AppPermissions {
   }
 
   AppPermissions._(this._excludePermissions, this._webtritCallkeepPermissions) {
-    _permissionsCache = ExpiringCache(
-      ttl: _cacheTTL,
-      compute: () => _fullPermissions.where((p) => !_excludePermissions().contains(p)).toList(),
-    );
+    _permissionsCache = ExpiringCache(ttl: _cacheTTL, compute: _computePermissions);
+  }
+
+  /// The non-excluded permissions to request. On web, permissions the platform
+  /// cannot handle are dropped - otherwise request()/status throw UnsupportedError.
+  List<Permission> _computePermissions() {
+    final excluded = _excludePermissions();
+    return _fullPermissions
+        .where((p) => !excluded.contains(p))
+        .where((p) => !kIsWeb || _webSupportedPermissions.contains(p))
+        .toList();
   }
 
   /// Manages permissions related to `webtrit_callkeep` plugin.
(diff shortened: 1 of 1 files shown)


lib/data/app_permissions.dart
```diff
@@ -196,10 +211,12 @@ class AppPermissions {
   }
 
   Future<bool> isContactPermissionGranted() async {
+    if (kIsWeb) return false; // contacts permission is not supported on web
```
SERDUN: **[code-review] (CONFIRMED) Web contacts path lands on a dead-end permission-failure UI.** `isContactPermissionGranted()` now returns `false` on web. In `LocalContactsSyncBloc._onStarted`, after `featureEnabled` and the agreement check pass, a `false` here emits `LocalContactsSyncPermissionFailure` -- the permission-denied screen whose "open settings" button is a no-op on web. So a web build with the local-contacts tab enabled (`ContactSourceType.local`) shows a dead-end error state instead of treating the feature as unavailable. (Reachable only when a web config enables local contacts; pre-PR this path threw UnsupportedError instead.) Consider emitting a feature-unavailable state on web rather than a permission failure.

lib/data/app_permissions.dart
```diff
@@ -18,6 +20,12 @@ class AppPermissions {
   /// The list of all possible permissions that the app may request.
   static const _fullPermissions = [..._requiredPermissions, Permission.camera, Permission.contacts, Permission.sms];
 
+  /// Permissions supported by permission_handler's web implementation. Requesting
+  /// or querying any other permission on web throws UnsupportedError, so on web
+  /// the requested set is restricted to these. (contacts/sms are not available on
+  /// web; camera + microphone cover the call use case.)
+  static const _webSupportedPermissions = [Permission.microphone, Permission.camera];
```
SERDUN, line 27: **[code-review] (PLAUSIBLE) `_webSupportedPermissions` is a hand-maintained parallel subset of `_fullPermissions`.** When a new call-relevant, web-capable permission is added to `_fullPermissions`, a maintainer must remember to also add it here -- forgetting silently drops it on web with no compile error, so the web build stops requesting a permission it actually supports. A derived expression or a single source of truth would avoid the drift.

lib/data/app_permissions.dart
```diff
@@ -196,10 +211,12 @@ class AppPermissions {
   }
 
   Future<bool> isContactPermissionGranted() async {
+    if (kIsWeb) return false; // contacts permission is not supported on web
     return isPermissionGranted(Permission.contacts);
   }
 
   Future<bool> requestContactPermission() async {
+    if (kIsWeb) return false; // contacts permission is not supported on web
```
SERDUN: **[code-review] (PLAUSIBLE) Contacts methods hardcode `if (kIsWeb) return false` -- a second source of truth.** Both `isContactPermissionGranted` and `requestContactPermission` encode "contacts unsupported on web" inline, while `_computePermissions` derives the same fact from `_webSupportedPermissions`. If web contacts support ever lands (allowlist updated), these inline returns keep contacts broken. Deriving from the allowlist (e.g. `if (!_webSupportedPermissions.contains(Permission.contacts)) return false;`) keeps one source of truth.

lib/data/app_permissions.dart
SERDUN, line 1: **[code-review] (PLAUSIBLE, app_permissions.dart:135) Required-only status path bypasses the web allowlist.** `getPermissionStatuses(onlyRequired: true)` reads `_requiredPermissions` directly, not the web-filtered `permissions` getter; the first loop of `isDenied` does the same. Today `_requiredPermissions == [microphone]` (web-supported), so it is safe by coincidence. If any non-web permission is ever added to `_requiredPermissions`, these paths call `.status` on web and throw `UnsupportedError` -- the exact failure this PR prevents elsewhere.

SERDUN, line 1: **[code-review] (PLAUSIBLE, app_permissions.dart:208) Public `isPermissionGranted` is unguarded.** External call sites (`call_active_scaffold.dart:153`, `main_shell.dart:513/563`) pass `Permission.camera` straight in; camera is in `_webSupportedPermissions` so it works, but this entry point is not covered by the allowlist guard. Passing any non-allowlisted `Permission` here on web throws `UnsupportedError` at the call site.

PR #1423, opened 2026-06-22
feat(web): allow overriding the session bundle_id on web

SERDUN:
## Overview

PR 4 of the decomposed web series. **Stacked on #1422** (base `feat/web-permissions`).

Web has no platform bundle identifier, so `packageInfo.packageName` resolves to the pubspec project name on web, which the backend rejects with `unconfigured_bundle_id` (blocks login + autoprovision).

## Scope
- Add `WEBTRIT_APP_WEB_BUNDLE_ID` dart-define + `EnvironmentConfig.resolveBundleId(...)`, used by the session (login) and autoprovision requests. On web it overrides the bundle_id; native always uses `packageInfo.packageName`.
- Warn once at bootstrap on web when the define is unset (otherwise login fails with no client-side hint).
- Document the define in `dart_define.json`.

Long-term proper fix is server-side: register the web bundle_id on PortaOne so the client override is unnecessary.

Verified: `flutter analyze` clean; `flutter build web` green.


dart_define.json
```diff
@@ -10,6 +10,8 @@
   "WEBTRIT_APP_LINK_DOMAIN": "app.webtrit.com",
   "__WEBTRIT_APP_LINK_DOMAIN_DESCRIPTION": "If provided, enables Android App Links integration via intent-filter. Must match assetlinks.json configuration on the target domain. This functionality is configured via Android flavors — if the value is non-empty, the 'deeplinks' flavor is selected automatically.",
   "WEBTRIT_APP_NAME": "WebTrit",
+  "WEBTRIT_APP_WEB_BUNDLE_ID": "com.webtrit.app",
```
SERDUN: **[code-review] (CONFIRMED) New `WEBTRIT_APP_WEB_BUNDLE_ID` ships ACTIVE with a hardcoded `com.webtrit.app` default.** Previously the sent bundle_id == this build`'`s `packageInfo.packageName`, so a white-label native build always sent its own brand id. On web there is no packageName, hence this override -- but shipping it as a non-underscored key defaulting to `com.webtrit.app` means a white-label web build that uses the stock `dart_define.json` without customizing this key silently sends `bundle_id=com.webtrit.app`. The backend then either rejects that brand`'`s login/autoprovision or authenticates the web session against the wrong (webtrit.app) registration -- wrong-tenant identity. Safer to ship it underscore-prefixed (`_WEBTRIT_APP_WEB_BUNDLE_ID`) with an empty default so each web brand opts in explicitly; a brand that forgets then fails loudly (packageName fallback -> unconfigured_bundle_id + the new bootstrap warning) instead of silently impersonating webtrit.app.

lib/environment_config.dart
```diff
@@ -52,6 +54,25 @@ class EnvironmentConfig {
...
+  // to the pubspec project name on web, which the backend rejects with
+  // `unconfigured_bundle_id`. This build-time override lets a web build send a
+  // bundle_id that is registered on the server for the `web` app type.
+  // TODO(web): the cleaner long-term fix is to register the web bundle_id on the
+  // server (PortaOne/Core) for the `web` app type, so this client override is no
+  // longer needed.
+  static const WEB_BUNDLE_ID__NAME = 'WEBTRIT_APP_WEB_BUNDLE_ID';
+  static const WEB_BUNDLE_ID = bool.hasEnvironment(WEB_BUNDLE_ID__NAME)
+      ? String.fromEnvironment(WEB_BUNDLE_ID__NAME)
+      : null;
```
SERDUN: **[code-review] (CONFIRMED) Empty-string override is treated as a valid bundle_id.** `bool.hasEnvironment` is true for an empty value, and `"" ?? packageName` keeps `""` (the `??` only catches null), so `resolveBundleId` returns `""` rather than falling back to `packageName`. An operator who disables the key by blanking its value in `dart_define.json` (instead of underscoring it -- mirroring how `_WEBTRIT_APP_ABOUT_URL: ""` looks in this file) makes web send `bundle_id=""`; the server rejects it with `unconfigured_bundle_id`, there is no packageName fallback, and the bootstrap warning never fires (`WEB_BUNDLE_ID` stays non-null). Normalize empty to null at the source (this also closes the warning-coverage gap below):
```suggestion
  static const _webBundleIdValue = String.fromEnvironment(WEB_BUNDLE_ID__NAME);
  static const WEB_BUNDLE_ID = _webBundleIdValue == '' ? null : _webBundleIdValue;
```

lib/bootstrap.dart
```diff
@@ -84,11 +84,19 @@ Future<InstanceRegistry> bootstrap() async {
     remoteDatasource: systemInfoRemoteDatasource,
   );
 
+  if (kIsWeb && EnvironmentConfig.WEB_BUNDLE_ID == null) {
```
SERDUN: **[code-review] (CONFIRMED) Warning is emitted before any log sink is attached.** This fires at bootstrap line ~88, but `AppLogger.init(...)..attachToLogger(Logger.root)` runs later in bootstrap (~line 158) and `main.dart` attaches its root listener only after `await bootstrap()` returns (main.dart:46). So on a misconfigured web build that reaches this branch, the warning ("falling back to packageName ... unconfigured_bundle_id") is dropped before any file/remote sink exists -- exactly the scenario it was added to explain is invisible in collected logs. Move the check after the logger is attached (or re-log it there).

SERDUN: **[code-review] (CONFIRMED) Missing-define warning guard is narrower than the failing input set.** The guard is `WEB_BUNDLE_ID == null`, but an empty-string override also yields a backend-rejected bundle_id while keeping `WEB_BUNDLE_ID` non-null -- so the diagnostic meant to replace the lost "packageName is always valid" guarantee does not fire for that case. Normalizing empty -> null in `WEB_BUNDLE_ID` (suggested on environment_config.dart) makes this guard cover it.

lib/environment_config.dart
```diff
@@ -52,6 +54,25 @@ class EnvironmentConfig {
...
+
+  /// The bundle_id sent to the backend in session/autoprovision requests.
+  ///
+  /// On web there is no platform bundle id, so [WEB_BUNDLE_ID] (the
+  /// WEBTRIT_APP_WEB_BUNDLE_ID dart-define) overrides [packageName] when set;
+  /// otherwise it falls back to [packageName]. Native always uses [packageName].
+  static String resolveBundleId(String packageName) => (kIsWeb ? WEB_BUNDLE_ID : null) ?? packageName;
```
SERDUN, line 75: **[code-review] (PLAUSIBLE) Web bundle_id is resolved per-consumer instead of at the source value.** `resolveBundleId()` must be called at every site that sends a bundle_id; the two current producers (bootstrap`'`s `appBundleId` and `AutoprovisionCubit._bundleId`) were both hand-edited identically. A future site that reads `packageInfo.packageName` directly will get the pubspec name on web and be rejected with `unconfigured_bundle_id`. The codebase already special-cases web at the data layer (e.g. `device_info.dart``'`s `kIsWeb` branch); resolving the web bundle id once in `package_info.dart``'`s `packageName` would remove the per-call-site footgun.

PR #1437, opened 2026-06-23
feat(app): support running embedded in a host (embeddedPreview + injection)

SERDUN:
## Overview

Let `webtrit_phone` run **embedded inside another Flutter host** — the theme configurator's realtime preview mounts the real app in-process. Two parts, both in `bootstrap.dart` / `main.dart` (RootApp) / `app/view/app.dart` (App):

1. **Firebase-free bootstrap.** `bootstrap({bool embeddedPreview = false})`: when embedded, skip Firebase app/messaging/local-push init, use `SharedPreferencesAppIdProvider` instead of the Installations-backed `FirebaseAppIdProvider`, and skip Firebase Remote Config. The host owns the default Firebase app (which may be misconfigured or absent — locally the configurator's app has an empty `projectId`), so any Firebase touch in the embedded app would throw.
2. **External config injection.** `RootApp`/`App` accept external `ThemeSettings` / `ThemeMode` / `FeatureAccess` streams (+ initial seeds) so the host drives the live appearance and feature/login/tab config. Theme is applied via `AppThemeSettingsChanged` / `AppThemeModeChanged` (no prefs write); the `FeatureAccess` `StreamProvider` source is swapped to the external stream. The analytics provider is made lazy and the analytics navigator observer is skipped when embedded, so `FirebaseAnalytics.instance` is never touched.

## Why

Part of the configurator realtime-preview epic (depends on the EnvironmentConfig runtime-override PR for the dart-define injection path the host uses). Standalone builds are unaffected.

## Notes

- Standalone behaviour byte-identical: `embeddedPreview` defaults to `false`, all external streams default to `null`.
- `flutter analyze lib` green.

### DIFF
--- a/lib/app/view/app.dart
+++ b/lib/app/view/app.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 
 import 'package:flutter_bloc/flutter_bloc.dart';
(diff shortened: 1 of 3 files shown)


SERDUN:
Splitting into two PRs: (1) embeddedPreview Firebase-free bootstrap, (2) external Theme/FeatureAccess injection.

PR #1438, opened 2026-06-23
feat(app): Firebase-free embeddedPreview bootstrap

SERDUN:
## Overview

Add an `embeddedPreview` mode so `webtrit_phone` can run **inside another Flutter host** (the theme configurator's realtime preview), where the host owns the default Firebase app — which may be misconfigured or absent (locally the configurator's app has an empty `projectId`), so any Firebase touch in the embedded app would throw.

- `bootstrap({bool embeddedPreview = false})`: when embedded, skip Firebase app/messaging/local-push init, use `SharedPreferencesAppIdProvider` instead of the Installations-backed `FirebaseAppIdProvider`, and skip Firebase Remote Config.
- `RootApp`/`App` take an `embeddedPreview` flag; the analytics provider becomes **lazy** (`create:` instead of `.value`) and the analytics navigator observer is skipped when embedded, so `FirebaseAnalytics.instance` is never touched.

## Why

First half of the embedded-host support for the realtime-preview epic (split from #1437). The second half — driving the live theme/feature config into the embedded app via external streams — is a follow-up PR stacked on this one.

## Notes

- Standalone behaviour is unchanged: `embeddedPreview` defaults to `false`.
- `flutter analyze lib` green.

### DIFF
--- a/lib/app/view/app.dart
+++ b/lib/app/view/app.dart
@@ -21,7 +21,13 @@ import 'package:webtrit_phone/resolvers/resolvers.dart';
 final _logger = Logger('AppWidget');
 
 class App extends StatefulWidget {
-  const App({super.key});
+  const App({super.key, this.embeddedPreview = false});
+
+  /// Whether the app runs embedded inside another Flutter host (realtime
+  /// preview). When true, Firebase-backed integrations such as the analytics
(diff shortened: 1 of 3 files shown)


SERDUN:
Closing — reworking the embedded-host support into a cleaner single abstraction.

PR #1441, opened 2026-06-24
feat(config): gate Firebase behind WEBTRIT_APP_FIREBASE_ENABLED

SERDUN:
## Overview

Adds a `WEBTRIT_APP_FIREBASE_ENABLED` config (default `true`, runtime-overridable via `EnvRegistry`) that lets the app run **Firebase-free** when a host owns the default Firebase app — specifically the theme configurator's realtime preview, which mounts this app in-process.

This is the first of two PRs split out of the embedding work (the second adds host theme / feature-config injection).

## Changes

- `environment_config.dart` — new `FIREBASE_ENABLED` getter (override-first, compile-time default `true`).
- `bootstrap.dart` — gate every Firebase touchpoint behind the flag:
  - Firebase app / messaging / local pushes init.
  - App-id provider: `FirebaseAppIdProvider` when on, `SharedPreferencesAppIdProvider` when off.
  - Remote Config: falls back to the local shared-preferences cache when off (and on init failure).
- `main.dart` — `AppAnalyticsRepository` provided lazily so `FirebaseAnalytics.instance` is only touched when the observer is attached.
- `app.dart` — analytics navigator observer attached only when the flag is on.
- `dart_define.json` / `docs/environment.md` — document the new flag.
- `test/environment_config_test.dart` — covers default-true and false-override.

## Verification

- `flutter analyze` — clean.
- `flutter test test/environment_config_test.dart` — 8/8 pass.
- Standalone builds default to `true` and behave exactly as before (no overrides applied).

### DIFF
--- a/dart_define.json
+++ b/dart_define.json
@@ -1,4 +1,6 @@
 {
+  "WEBTRIT_APP_FIREBASE_ENABLED": true,
+  "__WEBTRIT_APP_FIREBASE_ENABLED_DESCRIPTION": "Whether Firebase (app, messaging, local push, Remote Config, Installations, Analytics) is initialised. Default true. Disable when the app is embedded in a host that owns the default Firebase app (e.g. the theme configurator's realtime preview) so the embedded app runs Firebase-free.",
   "WEBTRIT_APP_DATABASE_LOG_STATEMENTS": false,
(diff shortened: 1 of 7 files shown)


SERDUN:
Closing to restart the embedding support under a single, consistent injection model.

PR #1442, opened 2026-06-24
feat(app): allow a host to inject theme and feature config

SERDUN:
## Overview

Lets an external host that mounts this app in-process — the theme configurator's realtime preview — supply the **theme** and **feature config** it should render, instead of the app's own bootstrap-built defaults.

This is the second of two PRs split out of the embedding work (the first, #1441, makes Firebase optional). The two are independent; the realtime preview consumes both.

## Changes

- `main.dart` — `RootApp` gains optional `themeSettings`, `themeMode` and `featureAccess` parameters, provided down the tree:
  - `FeatureAccess` — a host-supplied value replaces the reactive `StreamProvider`; otherwise the stream is used as before.
  - `ThemeSettings?` / `ThemeMode?` — provided as nullable values (null in a normal run).
- `app.dart` — when a non-null host theme is present, `App` pushes it into the `AppBloc` (`AppThemeSettingsChanged` / `AppThemeModeChanged`) in `didChangeDependencies`, so **`AppState` stays the single source of truth** for theme — no parallel override path, no streams smeared through the widget tree.

## Verification

- `flutter analyze` — clean.
- All parameters are null in a standalone run; behaviour is unchanged.

### DIFF
--- a/lib/app/view/app.dart
+++ b/lib/app/view/app.dart
@@ -31,6 +31,9 @@ class _AppState extends State<App> {
   late final AppBloc appBloc;
   late final AppRouter appRouter;
 
+  ThemeSettings? _lastHostThemeSettings;
+  ThemeMode? _lastHostThemeMode;
+
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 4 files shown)


SERDUN:
Closing in favour of a cleaner approach: RootApp as the single injection boundary, everything below consuming via inherited providers (theme lifted out of the AppState render path).

PR #1447, opened 2026-06-25
fix: pull video calls via configurable video strategy (WT-1618)

SERDUN:
## Overview

Call Pull of a video call failed: the pull offer was audio-only, so the server's
answer kept the original `m=video` and `setRemoteDescription` rejected it
("order of m-lines in answer doesn't match order in offer"), failing the pull.

The fix makes the pull media layout match the server's answer, and the behaviour
is selectable via a `CallPullVideoStrategy` so operators can pick a backend-free
default or a backend-assisted mode without a release.

### Strategies

Resolved as `Remote Config override > configurator default > softMute`:

- **softMute** (default, no backend dependency): video calls stay pullable. The
  pull offer carries a **recvonly** video m-line (reusing the caller's
  `PeerConnectionPolicyApplier`), so the offer/answer media layouts match. The
  camera is never opened by the pull, so an audio-only pull on a video dialog
  does not trip camera permission. Validated on PortaSwitch, where the
  dialog-info carries no media-type flag.
- **hideVideo** (backend-assisted): video dialogs are excluded from the pull list
  (`CallPullCubit`) and the pull offer omits the video m-line. This consumes the
  dialog-info `has_video` flag, so it needs a backend that reports media type.

### Wiring

- The configurator default lives in `SupportedFeature.callPull` (carried as a
  string in `webtrit_appearance_theme`; the app owns the `CallPullVideoStrategy`
  enum, and the `FeatureAccess` `CallMapper` bridges them - appearance-theme
  types stay out of the app runtime models).
- The Firebase Remote Config override key is `feature_call_pull_video_strategy`.

### `has_video` plumbing

Signaling -> `DialogInfo` -> drift + migration v24. Under **softMute** the fix
works with `has_video` absent/null; under **hideVideo** it is the gating signal.
This supersedes the "hide pull for video" workaround (#1443).

### Changes
- `peer_connection_policy_applier.dart`: strategy pattern -
  `VideoOfferStrategy { inactiveSendrecv, recvonly }`; recvonly adds a video
  transceiver without opening the camera.
- `call_bloc.dart`: for a pull (`fromReplaces != null`) under softMute, add the
  recvonly video m-line before `createOffer`.
- `call_pull_cubit.dart`: under hideVideo, drop known-video dialogs from the
  pull list.
- `feature_access` (models + mapper): resolve `CallPullVideoStrategy` from the
  Remote Config override, configurator default, then softMute.
- `dialog_info.dart`: `pullable` stays media-agnostic (the strategy decides
  downstream).

### Testing
Verified on PortaSwitch (testenv): audio and video pull both connect under
softMute, no `setRemoteDescription` m-line mismatch, camera stays off.
`flutter analyze` clean + full test suite green.


SERDUN:
Superseded by #1449 (re-opened as feat with the configurable Call Pull video strategy; same branch content, accurate title/description).

PR #1485, opened 2026-07-08
fix: require a backend-produced rejection to confirm a missed session (WT-1557)

SERDUN:
## Overview

Stacked on #1483. Code-review follow-up: the parent PR still treated ANY unmapped 4xx on the verification request as positive evidence of a dead session. That bucket is populated mostly by intermediaries, not by the backend: a rate-limiting load balancer answering 429 during an outage-recovery reconnect storm, a proxy 408, a WAF 403 with no backend error body. One such response was enough to force-logout a perfectly valid session - during exactly the kind of outage the parent PR is meant to survive (no retry at the API-client level for status responses, no debounce on the logout dispatch).

This PR narrows the verdict: a 4xx confirms a missed session only when the backend itself produced the rejection.

- status must not be transient by HTTP semantics (408 Request Timeout, 425 Too Early, 429 Too Many Requests - the server explicitly asks to retry)
- the response body must carry a backend error code (`errorCode`); status-only responses come from intermediaries

Everything else resolves to `SessionUnverifiable` (keep the session, the signaling layer keeps re-reporting and a later verification gets the definitive answer). Real backend verdicts are unaffected: the known session-death rejections arrive as typed exceptions handled earlier, and any future backend rejection code still confirms the logout under the new rule.

The HTTP semantics live in `webtrit_api` as classification getters on `RequestFailure` (`isClientError`, `isServerError`, `isTransient`, `errorCode`), so the session layer composes only the product policy and no raw status-code ranges leak outside the API package.

## Changes

- `webtrit_api`: classification getters on `RequestFailure` + 4 unit tests
- `SessionVerifier`: verdict condition narrowed to `isClientError && !isTransient && errorCode != null`; class doc updated
- `test/app/session/session_verifier_test.dart`: 4 new cases (429 with a code, 408, 425, status-only 403 -> all `SessionUnverifiable`)

## Testing

- `flutter analyze` clean
- `webtrit_api` package tests green (8), full app suite green, including 14 `SessionVerifier` tests

### DIFF
--- a/lib/app/session/session_verifier.dart
+++ b/lib/app/session/session_verifier.dart
@@ -44,11 +44,12 @@ class SessionUnverifiable extends SessionVerdict {
 
 /// Verifies over the REST API whether the account session is still valid.
 ///
-/// A logout requires positive evidence: transport failures and 5xx responses
-/// carry no verdict about the session and resolve to [SessionUnverifiable],
-/// so a backend outage never logs the user out. If the session is really
-/// gone, the signaling layer keeps reporting it and a later verification gets the
-/// definitive answer once the backend recovers.
+/// A logout requires positive evidence: transport failures, 5xx responses
+/// and 4xx responses produced by intermediaries (load balancers, proxies,
+/// WAFs) carry no verdict about the session and resolve to
+/// [SessionUnverifiable], so a backend outage never logs the user out. If
+/// the session is really gone, the signaling layer keeps reporting it and a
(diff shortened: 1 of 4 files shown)


SERDUN:
Superseded by a different ordering: the webtrit_api preparation (RequestFailure classification getters) goes in first as a standalone PR from develop; the verifier verdict change will follow on top of the base PR once that lands. Branch is kept for the follow-up.

PR #1499, opened 2026-07-10
feat: client-side voicemail transcription on a reusable media transcription engine (WT-719)

SERDUN:
## Overview

Voicemails can now be read, not only listened to: each voice message gets a text transcript rendered below the audio player, with clear "transcribing", "downloading transcription model" and "unavailable" states. Transcription is built as a media-agnostic subsystem - voicemail is its first consumer, and call recordings or chat voice notes can plug into the same pipeline later without schema or engine changes.

Transcripts are produced by Whisper behind an engine-agnostic interface, with two switchable sources:

- **local** - on-device Whisper (`whisper_ggml` / whisper.cpp): the ggml model is downloaded on first use or right when the user picks a tier; audio is converted to 16 kHz mono WAV (telephony voicemail is 8 kHz, sometimes ulaw) by the bundled ffmpeg before recognition. Not available on web (FFI), guarded by a conditional import.
- **remote** - any OpenAI-compatible speech-to-text endpoint (`POST .../audio/transcriptions`): OpenAI, a self-hosted Speaches/faster-whisper server, etc. Accepts a full endpoint URL or its `/v1` base, optional Bearer key.

The source is selected per brand in the white-label app config (`assets/themes/app.config.json`, new top-level `transcription` section - transcription is media-agnostic, voicemail is just its first consumer). The stock config ships with the local engine enabled; configs without the section fall back to the disabled schema default:

```json
"transcription": {
  "mode": "local | remote | disabled",
  "language": "optional ISO 639-1 hint; omit to auto-detect",
  "local": { "model": "base" },
  "remote": { "url": "https://stt.example.com/v1", "apiKey": "optional", "model": "whisper-1" }
}
```

Note: the configurator currently edits the old `voicemail.transcription` path and needs a follow-up PR for the new section.

## Architecture

- **`packages/data/app_transcription`** is self-contained and storage-agnostic:
  - `TranscriptionDataSource` contract (bytes -> text, `engine` id, `downloadState`, `prepareEngine`, dispose) with the local (io + web stub) and remote implementations; the HTTP wiring stays internal to the package. `TranscriptionConfig` and `createTranscriptionDataSource` live here too - config -> engine mapping is the package's own concern.
  - `TranscriptionService` - a fire-and-forget pool provided session-wide in the main shell. Consumers hand media off through the narrow `MediaTranscriber` contract (enqueue + forget + model switch) and never await a result: the pool processes its queue with per-engine concurrency (sequential for the compute-bound local engine, 3-wide for the network-bound remote one), dedups queued/in-flight items, marks every queued item in progress immediately (after a model switch the whole backlog shows "transcribing" instead of a blank list) and attributes every result to the engine that produced it.
  - `TranscriptionStore` - the pool's storage delegate. The pool knows nothing about the database; every lifecycle fact goes to this interface.
- **The database is the single feedback channel.** The app implements the store as `TranscriptionStoreDriftImpl` over a new media-agnostic `transcriptions` table (schema v26, keyed by `(media_type, media_id)` with transcript, status and producing engine). The voicemail list joins that table; a watch over voice messages missing a transcription row drives re-enqueue (the pool's own lifecycle writes take rows out of the watch, so it cannot loop), and the store is consulted before work starts, so a raced re-enqueue cannot clobber a finished transcript.
- **Model choice** lives in a session-wide model service: it persists the preference first and switches the pool after, reverting the override best-effort when the switch is refused (a disposed pool throws instead of silently no-opping), and serializes rapid re-selection. A real engine change wipes the stored transcriptions and regenerates every message; re-picking the same tier, or any pick in the remote mode, is a no-op. The config's `local.model` only sets the predefined default tier.
- **Switch safety.** Enqueues landing mid-switch are dropped and re-fed by the database watch, so a fresh item can never be transcribed and persisted by the outgoing engine; in-flight work of the old engine is invalidated by generation and identity guards, with staleness re-checked between the download and the inference so dead work is dropped early.

## UX

- `VoicemailTile`: transcript below the player (collapsed to 3 lines, tap to expand), progress row with a spinner that honestly distinguishes "downloading transcription model" from "transcribing", italic unavailable note.
- All voicemail toolbar actions live in a single trailing three-dot menu: transcription model (shown only when a local engine is actually running - a functional gate, not a config flag), storage and cache, and delete, with the multi-selection counter badge on the menu button.
- The model choice is a dedicated transcription settings page (same pattern as the cache management page): Fast (base, ~142 MB), Balanced (small, ~466 MB), Accurate (medium, ~1.5 GB), plus the brand default when it is a different tier. Tiers already on disk carry a "Downloaded" badge; picking a tier starts its download immediately with a live progress bar (progress is exact - the package streams the file itself - and throttled to ~1% steps) and a retry action after a failure.
- The downloaded model files are a section of the Storage & cache page: their total size is shown (stray partial downloads included) and they can be cleared in one tap - the active model is downloaded again on the next transcription (readiness is re-verified against the disk, not cached).

## Robustness

- Transcription starts only after the server confirms voicemail support, so cached messages are never marked terminally unavailable by doomed downloads; transiently failed messages retry even after failed refreshes; terminal failures are not retried automatically.
- Refetches preserve stored transcripts (conflict-limited upsert) and never re-transcribe; fax messages are skipped; orphan transcription rows are swept after each fetch; deleted media loses its rows even when the deletion races session teardown.
- 401 during transcription rolls the item back and delegates to the session guard; a failed transcript write is retried later instead of being classified terminal; a failed wipe keeps the current engine and surfaces the error; an aborted model download cleans up its partial file.
- The repository disposes its database watches with the session, the pool ignores calls after dispose, and the voicemail list rebuilds only when the model-downloading state actually flips, not on every progress tick.

## Trade-offs

- `whisper_ggml` pulls `ffmpeg_kit_flutter_new_min`, which adds app size; accepted for v1 because it makes audio-format handling robust. The engine sits behind the interface, so swapping it (e.g. for sherpa-onnx) or moving fully to remote is a single-class change.
- Transcription runs while the app is alive (triggered by voicemail fetch/refresh); there is no background push pipeline for voicemails.

## Testing

- Unit and widget coverage across the stack: the datasources (remote request shape, endpoint resolution, auth, language, error mapping; local model download validation, progress states, aborted-download cleanup, cache sizing/deletion and self-healing re-download), config parsing and the factory, the pool (concurrency, queued-item visibility, mid-switch enqueue drop, disposed-pool semantics, download-state mirroring across switches), the pool via the repository (success, failures transient/terminal, refetch idempotency, no-retry after unavailable, fax skip, model swap and regeneration, mid-flight switch discard, deleted-message race, unsupported-server gate), the model service (persist-and-revert, disposed pool), deterministic contact resolution, the transcription settings page (options, badges, progress, retry, rollback-to-actual), the overflow menu, the models cache section and migration v26 data integrity.
- Full suite green, `flutter analyze` clean (app + packages).
- Real-device run of the local engine (model download + FFI build) is still pending - that is the next verification step while the PR is in draft.

## Documentation

`docs/features/transcription.md` documents the whole subsystem (architecture, contracts, consumer rules, a guide for adding new media consumers); `docs/application_config.md` gained the `transcription` section.

## Related PRs

Standalone changes were extracted to keep this diff on topic: #1507 (format debt), #1508 (deterministic voicemail contact resolution), #1509 (binary attachment downloads + `file_format` forwarding, which this branch depends on), #1510 (explicit trusted certificates import) - all merged to develop and rebased out of this diff. #1511 (model download status and progress) was developed as a stacked PR and is merged into this branch.


SERDUN:
image

digiboridev (approved):

PR #1518, opened 2026-07-16
fix(signaling): skip an unparseable call_log entry instead of failing the handshake

SERDUN:
## Overview

WT-1729. Stacked on #1517 (the narrow transfer_accepted/transfer_failed fix). This is the broader hardening step that PR called out as deliberately out of scope: `StateHandshake.fromJson` throws for any `call_logs` entry it cannot parse (any event type this client version does not recognize). Since the state handshake is replayed on every reconnect, one bad entry doesn't just fail once - it fails identically on every subsequent reconnect attempt for as long as the call carrying it stays alive. This was the actual mechanism behind the transfer_failed crash-loop before #1517 taught the client to recognize that specific event; this PR makes the mechanism itself resilient, so the same thing can't happen again for any other future/unrecognized event type.

## Changes

- Extracted the near-identical main-line/guest-line `call_logs` parsing (previously duplicated inline, ~25 lines each) into one `_parseCallLogs` helper - net code reduction alongside the fix.
- Each `call_logs` entry's parse is now wrapped in try/catch: an unparseable entry is logged and skipped, not thrown - it only loses that one log entry instead of failing the whole handshake (and, on reconnect, doing that again forever).

## Deliberately out of scope

Only the `call_logs` replay path in `webtrit_signaling` (`state_handshake.dart`) is touched - this is what turns a single bad event into a *loop*, which is the actual "storm" symptom. The live single-event path (`webtrit_signaling_client.dart` `_onMessage`) still treats an unrecognized event as fatal for a single occurrence; changing that too would be a broader behavioral call (does the app want to silently ignore any future unrecognized live event, or is treating it as a connection problem the right signal to catch protocol drift quickly) - not made here.

## Testing

- New test: a `call_logs` array with one valid entry and one entry with an unrecognized event type - `StateHandshake.fromJson` returns normally and the result contains only the valid entry. Verified this test fails against the pre-fix code (temporarily swapped in the base branch's version of the file) with the exact `ArgumentError: Unknown call event type` that caused the original crash-loop.
- `webtrit_signaling`: `flutter analyze` clean, `dart test` 104/106 (2 pre-existing unrelated failures in `line_error_event_test.dart`, confirmed present on the base branch too).
- `webtrit_phone`: `flutter analyze` clean, full pre-push suite 1134/1134 green.

### DIFF
--- a/packages/webtrit_signaling/lib/src/handshakes/state_handshake.dart
+++ b/packages/webtrit_signaling/lib/src/handshakes/state_handshake.dart
@@ -1,9 +1,12 @@
 import 'package:equatable/equatable.dart';
+import 'package:logging/logging.dart';
 
 import '../events/events.dart';
 import '../requests/requests.dart';
 import '../responses/responses.dart';
 import 'handshake.dart';
+
+final _logger = Logger('StateHandshake');
 
 enum RegistrationStatus {
   registering,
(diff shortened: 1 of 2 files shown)


SERDUN:
Closing for now - this PR's approach (silently drop an unparseable call_logs entry) surfaced a real design problem during ultra-review: it breaks the newest-first completeness invariant that handshake_processor.dart and isolate_manager.dart rely on via callLogs.firstOrNull to find the latest call event, and the bare catch also swallows genuine parsing bugs in already-known event types, not just the intended forward-compat case.

The reported symptom (transfer_accepted/transfer_failed causing signaling disconnects) is fully fixed by #1517 alone - this PR was only the broader, forward-looking hardening for any *future* unrecognized call_logs event, not required to close that.

Full chain and what a correct fix needs is written up in #1517's description, plus a TODO comment at the exact code it applies to (state_handshake.dart). Branch kept (not deleted) - the timestamp/payload-cast fix here (moving it inside the try/catch) is still a valid, standalone correctness fix worth cherry-picking whenever this is picked back up.

PR #1547, opened 2026-08-06
build: bump app version to 1.16.2+1 (WT-1781)

SERDUN:
## Overview

Patch for the release/1.16.2 line: brings in the app bar fix already merged to develop in #1544.

Setting the top bar Foreground color in the theme only affected the back button - screen titles, the connection status line and several action icons kept their own colors, while the profile avatar got tinted instead. With this patch one Foreground option colors the whole top bar and the avatar keeps its own palette, every screen's bar resolves through the same cascade (per-page override in the theme, then the global bar configuration, then the color scheme defaults), and the frosted overlay is applied only where the bar actually lets content show through. Themes that configure nothing new keep their current look.

## Adaptation for this line

The develop version of the change sits on top of the edge-to-edge rework, which is not part of 1.16.2. The system overlay handling is therefore left exactly as it is on this line, and the status line and avatar changes are applied to the top bar this line still uses. Everything else is taken over unchanged.

## Testing

- `flutter analyze` clean, full suite green (1152 tests) on this branch.
- Code generation re-run: the generated theme files match the sources.


SERDUN:
Superseded: the patch will be split into separate PRs on the release line instead of one adapted cherry-pick.

PR #1550, opened 2026-08-06
fix: recover a registration that keeps failing without a relogin (WT-1766)

SERDUN:
## Overview

When the phone system an account registers to is changed on the operator's side
(a different address, different credentials), the session the app is attached to
keeps trying the old ones and never registers again. The app sits on
"Waiting for connection..." indefinitely, and reconnecting does not help,
because it re-attaches to that very session. Until now the only way out was for
the user to log out and log back in.

The app now watches how long registration has been failing. Once it has been
broken for a few minutes, the app asks the server to build the session from
scratch instead of reattaching to it, so the current settings are picked up and
the account registers again on its own.

Three things keep this from making a bad situation worse:

- a call in progress always wins - the attempt is postponed until the call ends,
  since rebuilding the session would drop the call;
- losing the network does not count as a registration failure, so a tunnel or a
  flaky connection does not trigger anything;
- the number of attempts per outage is capped, so an outage affecting the whole
  phone system cannot turn every client into a source of extra load.

## Verification

- Reproduced on a local stack: the account's phone system address was changed
  underneath a running app. Before the change the app stayed stuck until a
  relogin; now it recovers on its own within the expected window.
- Unit tests cover the timing (fires after the threshold, ignores short
  failures, a recovery cancels it), the safety rules (an active call postpones
  it, offline time does not age the clock, the attempt budget is spent once),
  and that the request actually reaches the server as expected.
- Full analyze plus the root and package test suites pass.

WT-1766

### DIFF
--- a/docs/signaling_architecture_target.md
+++ b/docs/signaling_architecture_target.md
@@ -226,8 +226,36 @@ SignalingReconnectController({
   void Function(bool isAvailable)? onConnectionPresenceChanged,
   int notifyAfterConsecutiveFailures = 2,  // notify after this many consecutive failures
   bool reconnectEnabled = true,
+  Duration rebuildSessionAfter = const Duration(minutes: 5),  // see "Recovering a stuck registration"
+  int maxSessionRebuildsPerEpisode = 2,
 })
 ```
+
+### Recovering a stuck registration
+
+The server keeps one session per app and reads the SIP address and credentials
+into it once, when it creates it. If those change on the operator's side, the
+session keeps registering with the old ones and never succeeds again — and
+reconnecting does not help, because the socket re-attaches to that same session.
+Until now the only way out was for the user to log out and back in.
+
+The controller watches the registration status carried by the handshake and the
+session events. Once a failure has lasted `rebuildSessionAfter`, it schedules a
+reconnect itself — nothing else would, since a failed registration leaves the
+socket up — and that connect is made with `reregister: true`, which asks the
+server to build the session from scratch instead of attaching to it.
+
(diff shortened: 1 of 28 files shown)


steamingscut-dev:
Closing without merging.

Further investigation of the incident this was built for showed the customer's account was never actually stuck: their phone system was down for the whole captured period, the app and the server both kept retrying correctly, and the account would have recovered on its own once the platform came back. Relogin only appeared to help because it restarts the retry sequence immediately instead of waiting out the backoff.

That makes this change unnecessary here, and in that scenario actively unhelpful: it would ask the server to rebuild a perfectly valid session while the phone system is already struggling, adding load for no benefit.

The case it does solve - an account whose phone system address or credentials change underneath a live session - is real and reproducible, but it is better handled on the server, where it is fixed once for every client instead of only for upgraded ones. The branch stays in place should we need it.

WT-1766

SERDUN:
Closing without merging.

Further investigation of the incident this was built for showed the customer's account was never actually stuck: their phone system was down for the whole captured period, the app and the server both kept retrying correctly, and the account would have recovered on its own once the platform came back. Relogin only appeared to help because it restarts the retry sequence immediately instead of waiting out the backoff.

That makes this change unnecessary here, and in that scenario actively unhelpful: it would ask the server to rebuild a perfectly valid session while the phone system is already struggling, adding load for no benefit.

The case it does solve - an account whose phone system address or credentials change underneath a live session - is real and reproducible, but it is better handled on the server, where it is fixed once for every client instead of only for upgraded ones. The branch stays in place should we need it.

WT-1766

